PR #38, opened 2023-01-13
Add maf

deepchatterjeeligo:


libs/architectures/mlpe/architectures/embeddings/convolution.py
```diff
@@ -102,3 +102,39 @@ def get_flattened_size(self, size: int):
         out = self.downsampler(x)
         out = torch.flatten(out)
         return len(out)
+
+
+class NaiveEmbedding(torch.nn.Module):
```
EthanMarx, line 107: This is great. Can we put this in its own file and maybe call the class something like `FullyConnectedEmbedding` or `DenseEmbedding`?

libs/architectures/mlpe/architectures/flows/maf.py
```diff
@@ -0,0 +1,88 @@
...
+from nflows.transforms.autoregressive import (
+    MaskedAffineAutoregressiveTransform,
+)
+
+
+@dataclass
+class MaskedAutoRegressiveFlow(NormalizingFlow):
+    shape: Tuple[int, int]
+    num_transforms: int = 5
+    num_embedding_dim: Optional[int] = None
```
EthanMarx, line 19: Instead of assuming we are using the `NaiveEmbedding`, can we make this argument `embedding_net: torch.nn.Module = Flattener()` (See the coupling flow) ? That way, its flexible enough to take any embedding net, which is fundamentally separate form the specifics of the flow implementation

libs/architectures/mlpe/architectures/flows/maf.py
```diff
@@ -0,0 +1,88 @@
...
+                num_blocks=self.num_blocks,
+                activation=self.activation,
+                use_batch_norm=self.use_batch_norm,
+                use_residual_blocks=self.use_residual_blocks,
+            ),
+            RandomPermutation(features=self.param_dim),
+        ]
+        return single_block
+
+    def linear_block(self):
```
EthanMarx, line 67: Given this, it might make sense to remove the `@abstractmethod` from the base class. Users who wan't to create a flow with a linear method should then add that in the `transform_block` method.

libs/architectures/mlpe/architectures/flows/maf.py
```diff
@@ -0,0 +1,88 @@
...
+        # this is not needed for a MAF
+        return []
+
+    def distribution(self):
+        """Returns the base distribution for the flow"""
+        return StandardNormal([self.param_dim])
+
+    @property
```
EthanMarx, line 75: I like this. I think this should be a `property` of the ABC and the `build_flow` method should be an `abstractmethod` that subclasses implement themselves.

EthanMarx (changes requested):
This is awesome thanks for implementing this. A couple of structural changes but otherwise looks great.

libs/architectures/mlpe/architectures/embeddings/convolution.py
deepchatterjeeligo: Done

libs/architectures/mlpe/architectures/flows/maf.py
deepchatterjeeligo: I have set a default, But this is a little tricky because to make this dynamic `DenseEmbedding` there has to be several features of the embedding that also needs to be specified. In fact, I was thinking of inferring parts of this from the input data itself.

libs/architectures/mlpe/architectures/flows/maf.py
deepchatterjeeligo: done

libs/architectures/mlpe/architectures/flows/maf.py
deepchatterjeeligo: done

libs/architectures/mlpe/architectures/flows/maf.py
EthanMarx: Got it. My thought was we instantiate whatever Embedding network we wan't _outside_ of the flow class (possibly by inferring shape parameters from data as you mentioned), and then we can pass it directly to the flow. Let me know your thoughts

deepchatterjeeligo: Ummm...there are other details which need to be changed. For example, `num_embedding_dim` does not make sense of the embedding_net is a `Flattener`. One idea is to add arguments, `embedding_net_name` (this is a string value that can be loaded from the `embeddings` module), `embedding_net_args` and `embedding_net_kwargs`. But it will be hard to generalize due to other constraints. For example, the DenseEmbedding has to have the first matrix (layer) as a `self.shape[1] x <num_embedding_dim>` to match shapes etc. Some other embedding will need something else. So there has to be a case by case conditional for each embedding layer.

deepchatterjeeligo: So for now, I could do the above but raise a `NotImplemented` is `embedding_net_name` is not `Flattener` or `DenseEmbedding`

EthanMarx: So I don't think we have to do any of that with some cool functionality in `typeo`:

If we annotate the `MaskedAutoRegressiveFlow` and `CouplingFlow` as such (we'll have to remove their `dataclass` status which is fine)

```python
class CouplingFlow():
    def __init__(
    self, 
    shape: Tuple[int, int],
    embedding_net: Callable = Flattener(),
    ...
    **embedding_net_kwargs
):
        embedding_net = embedding_net(**embedding_net_kwargs)

```

then in the sandbox `pyproject.toml` we can do the following:

```toml
coupling = {num_flow_steps = 5, embedding_net = "mlpe.architectures.embeddings.DenseEmbedding", num_embedding_dims = 5}
```

And `typeo` will do the magic of loading in the `DenseEmbedding` module from the specified path. This way we can keep the flow implementations completely agnostic to the embedding networks. Also, if add another `embedding_net` we won't have to add it to the `embedding_net_name` parser.

(would love thoughts / confirmation from @alecgunny that this will work as expected)

deepchatterjeeligo: I support moving away from the dataclass, but that's a refactoring that can be done later. I will implement the embedding_net and embeddeding_net_kwargs and finish this patch

EthanMarx (approved):
Awesome. looks good

PR #39, opened 2023-01-19
get optimizer and scheduler as arguments to train

deepchatterjeeligo:
It is helpful to pass in different optimizer and scheduler names are arguments, so that they can be stored and passed from config in pyproject.toml

### DIFF
--- a/libs/trainer/mlpe/trainer/trainer.py
+++ b/libs/trainer/mlpe/trainer/trainer.py
@@ -110,14 +110,15 @@ def train(
     max_epochs: int = 40,
     init_weights: Optional[Path] = None,
     lr: float = 1e-3,
-    min_lr: float = 1e-5,
-    decay_steps: int = 10000,
-    weight_decay: float = 0.0,
     early_stop: int = 20,
     # misc params
     device: Optional[str] = None,
     use_amp: bool = False,
     profile: bool = False,
+    optimizer_fn: Callable = torch.optim.Adam,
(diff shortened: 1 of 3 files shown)


libs/trainer/mlpe/trainer/trainer.py
```diff
@@ -110,14 +110,15 @@ def train(
...
     lr: float = 1e-3,
-    min_lr: float = 1e-5,
-    decay_steps: int = 10000,
-    weight_decay: float = 0.0,
     early_stop: int = 20,
     # misc params
     device: Optional[str] = None,
     use_amp: bool = False,
     profile: bool = False,
+    optimizer_name: str = "Adam",
```
EthanMarx, line 118: Similar comments as #38. We can annotate this as 

```python
def train(
    optimizer: Callable, 
    optimizer_kwargs: Dict,
    ...
):
    optimizer = optimzer(optimizer_kwargs)
```
and then in the pyproject.toml have

```toml
[tool.typeo.scripts.train]
optimizer = "torch.optimizers.lr_scheduler.Adam"
optimizer_kwargs = {arg1: "arg1"}

deepchatterjeeligo: good idea.

deepchatterjeeligo: done.

EthanMarx (approved):
Great

PR #40, opened 2023-01-19
update CONDA PREFIX env variable in workflows

EthanMarx:


.github/workflows/unit-tests.yaml
```diff
@@ -60,16 +59,27 @@ jobs:
...
         run: |
-            if [[ "${{matrix.library}}" == "libs/architectures/" ]]; then
-                pinto -p $test_dir build -E wrapper
+            if [[ "${{ matrix.library }}" == "libs/architectures/" ]]; then
+                FLAGS="-E wrapper"
+            else
+                FLAGS=""
             fi
-            pinto -p $test_dir run pytest $test_dir/tests ${{matrix.args}}
+            echo $FLAGS
```
deepchatterjeeligo, line 75: Would you add a `printenv` somewhere here. So that next time it is easier to see the env vars.

EthanMarx: good idea done

deepchatterjeeligo (approved):

Issue #45, opened 2023-01-24
Remove jupyter notebook 


EthanMarx:
There is a unnecessary jupyter notebook in the architectures library that should be removed.

deepchatterjeeligo:
Closing this since notebook has been removed.

Issue #43, opened 2023-01-24
NormalizingFlow base class updates


EthanMarx:
To generalize to arbitrary `embeddings`, the NormalizingFlow class will need some small tweaks:

The `__init__` should look as follows:
```python
def __init__(self, param_dim: int, n_ifos: int, strain_dim: int, num_flow_steps: int, embedding_net: torch.nn.Module = Flattener())
    self.param_dim = param_dim
    self.strain_dim = strain_dim
    self.n_ifos = n_ifos
    self.num_flow_steps = num_flow_steps
    self.embedding_net = embedding_net
```
`n_ifos` will be used to infer the context dimension as described in #42.

This change will require all sub classes (e.g maf, coupling) to take as their `shape`a `Tuple[int, int, int]` to account for the new `n_ifos` argument

deepchatterjeeligo:
closed by #46

Issue #42, opened 2023-01-24
Function that infers `context_dim` from embedding


EthanMarx:
Add to the base `NormalizingFlow` class a `@property` that infers the shape of the context dimension.

```python
@property
def context_dim(self):
    # one as dummy batch dimension
    tensor = torch.randn(1, self.n_ifos, self.strain_dim)
    context_dim = self.embedding_net(tensor).shape[-1]
    return context_dim
```

PR #46, opened 2023-01-24
Flow base class updates

EthanMarx:
1. Removes torch deps from conda, and adds them to the `train` project `pyproject.toml`
2. Removes unneeded jupyter notebook (#45)
3. Adds context dim inference #42 to flow base class
4. Adds `n_ifos` as shape parameter to architectures #43 
5. Updates `trainer.py` to accomodate the above


libs/architectures/mlpe/architectures/flows/flow.py
```diff
@@ -32,10 +50,16 @@ def linear_block(self) -> transforms.Transform:
     def distribution(self) -> distributions.Distribution:
         pass
 
+    @property
+    def context_dim(self):
```
deepchatterjeeligo, line 54: let's keep `sample_tensor` as an optional argument with a default of what you have. Sometimes it is useful for debugging

libs/architectures/mlpe/architectures/flows/flow.py
```diff
@@ -32,10 +50,16 @@ def linear_block(self) -> transforms.Transform:
...
+        sample_tensor = torch.zeros((1, self.n_ifos, self.strain_dim))
```
deepchatterjeeligo, line 55: Will this respect device change?

libs/trainer/mlpe/trainer/trainer.py
```diff
@@ -110,15 +110,14 @@ def train(
...
+    weight_decay: float = 0.0,
+    early_stop: Optional[int] = None,
     # misc params
     device: Optional[str] = None,
     use_amp: bool = False,
     profile: bool = False,
-    optimizer_fn: Callable = torch.optim.Adam,
-    optimizer_kwargs: dict = dict(weight_decay=0),
-    scheduler_fn: Callable = torch.optim.lr_scheduler.CosineAnnealingLR,
-    scheduler_kwargs: dict = dict(eta_min=1e-5, T_max=10000),
```
deepchatterjeeligo, line 121: Why take this out? Let's just remove the lines from the project config now until the typeo trick is discovered.

libs/trainer/mlpe/trainer/trainer.py
```diff
@@ -191,15 +190,15 @@ def train(
         f["parameters"] = parameters.cpu().numpy()
 
     param_dim = parameters.shape[-1]
-    context_dim = strain.shape[-1] * strain.shape[-2]
+    _, n_ifos, strain_dim = strain.shape
```
deepchatterjeeligo, line 193:
```suggestion
    param_dim, n_ifos, strain_dim = strain.shape
```

libs/trainer/mlpe/trainer/trainer.py
```diff
@@ -191,15 +190,15 @@ def train(
         f["parameters"] = parameters.cpu().numpy()
 
     param_dim = parameters.shape[-1]
```
deepchatterjeeligo, line 192:
```suggestion

```

libs/trainer/mlpe/trainer/trainer.py
```diff
@@ -222,8 +221,13 @@ def train(
     logging.info(flow)
     logging.info("Initializing loss and optimizer")
 
-    optimizer = optimizer_fn(flow.parameters(), lr=lr, **optimizer_kwargs)
```
deepchatterjeeligo, line 225: Let's not revert this bit

projects/sandbox/pyproject.toml
```diff
@@ -72,10 +73,8 @@ batches_per_epoch = 200
 max_epochs = 300
 
 lr = 4e-3
-min_lr = 1e-5
-decay_steps = 10000
-weight_decay = 1e-5
-early_stop = 50
+
+early_stop = 100
```
deepchatterjeeligo, line 77: why this change?

libs/trainer/mlpe/trainer/trainer.py
EthanMarx: Would unfortunately require additional `typeo` development as far as I'm aware. For now let's stick with the hard coded values(feel free to change it to exponential, or whatever you want).

libs/trainer/mlpe/trainer/trainer.py
EthanMarx: No, so the parameters are separate. That dimension is the "batch" dimension (which we don't use)

libs/trainer/mlpe/trainer/trainer.py
EthanMarx: See above comment

projects/sandbox/pyproject.toml
EthanMarx: To let the network train for longer. Feel free to change. Like I said before, the parameters that are version controlled are more or less irrelevant until we find a configuration that works

libs/trainer/mlpe/trainer/trainer.py
```diff
@@ -290,14 +294,18 @@ def train(
...
                 since_last_improvement += 1
+
                 if since_last_improvement >= early_stop:
                     logging.info(
                         "No improvement in validation loss in {} "
                         "epochs, halting training early".format(early_stop)
                     )
                     break
+            else:
+                continue
```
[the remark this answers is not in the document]
EthanMarx: Is it standard to remove the `else:` entirely, or should I replace the `continue` with `pass`?

libs/trainer/mlpe/trainer/trainer.py
EthanMarx: I think we should for clarity (this is what we have in BBHnet). Given the optimizer is not our limiting factor, I think we can hold off on this until we have a better solution

libs/architectures/mlpe/architectures/flows/flow.py
EthanMarx: Good idea

libs/architectures/mlpe/architectures/flows/flow.py
EthanMarx: Not sure. Haven't thought much about device logic, but is definitely necessary

libs/architectures/mlpe/architectures/flows/flow.py
EthanMarx: Actually on second thought, if someone needs to debug, they can simply instantiate the `embedding` and pass tensors through there. I think this should stay as a `property` of the instantiated flow.

libs/architectures/mlpe/architectures/flows/flow.py
deepchatterjeeligo: well sample_tensor is on the default cpu device. If the flow is on cuda, this is expected to break.

libs/trainer/mlpe/trainer/trainer.py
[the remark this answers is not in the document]
deepchatterjeeligo: If you are thinking syntactic reason, we don't need an else if we have an elif

libs/trainer/mlpe/trainer/trainer.py
deepchatterjeeligo: It does not break anything right now.

deepchatterjeeligo: You just need to not mention it in the config

libs/trainer/mlpe/trainer/trainer.py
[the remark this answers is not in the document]
EthanMarx: Got it. I realized my logic is wrong anyways. Fixed

libs/trainer/mlpe/trainer/trainer.py
EthanMarx: Yeah, unfortunately I think we'll have to do this by hand for now given `typeo` constraints

projects/sandbox/pyproject.toml
deepchatterjeeligo: OK. Fine with me.

libs/trainer/mlpe/trainer/trainer.py
EthanMarx: Only because `typeo` is using the specified default. If you were to pass it as a config in the `pyproject.toml,` it would break. Since using the default is essentially equivalent to hard coding, lets just do that for now

libs/architectures/mlpe/architectures/flows/flow.py
EthanMarx: Let me write some tests for this

libs/trainer/mlpe/trainer/trainer.py
EthanMarx: I guess my philosophy is that the everything in the production repo should work as expected. As a hack in a branch this is of course totally fine.

EthanMarx:
Another addition: 
Updates `ml4gw` and uses `ml4gw` `distributions` module

libs/trainer/mlpe/trainer/trainer.py
deepchatterjeeligo: your philosophy is respected even without this change, and requires no breaking change in terms of typeo.

libs/trainer/mlpe/trainer/trainer.py
deepchatterjeeligo: I fail to understand what the constraint is.

deepchatterjeeligo:
Is the following the only issue BTW? That is dictionary arguments like `--scheduler-kwargs` cannot be passed like this? Because this is easily resolved by reading the keyword args as a string and running `eval`. While not general, it will work for the `scheduler_kwargs` argument in our `train` function.

```
train --background-path /home/deep.chatterjee/pe-base-dir/background.h5  \
        --waveform-dataset /home/deep.chatterjee/pe-base-dir/signals.h5 \
        --logdir /home/deep.chatterjee/pe-base-dir/log \
        --inference-params ["frequency", "quality", "hrss", "dec", "psi", "phi"] \
        --ifos ["H1", "L1"] \
        --sample-rate 2048 \
        --trigger-distance 0 \
        --kernel-length 2 \
        --fduration 1 \
        --highpass 32 \
        --batches-per-epoch 200 \
        --batch-size 512 \
        --device "cuda" \
        --max-epochs 100 \
        --init-weights /home/deep.chatterjee/pe-base-dir/training \
        --scheduler-fn "torch.optim.lr_scheduler.ExponentialLR" \
        --scheduler-kwargs "{gamma = 0.999}" \
        --verbose \
        maf \
        --num-transforms 10 \
```

fails with
```
main: error: argument --scheduler-kwargs: invalid dict value: '{gamma = 0.999}'
```

I want this flexibility since I am getting good convergence running with multiple schedules (at least until we build in the feature of a scheduler cycle which does all of this from the config). At least a shell script with multiple train lines with different schedulers with achieve this. Whatever solution gets to this point I am happy to work with.

EthanMarx:
Yes the basic problem is that for Dictionary arguments, `typeo` can currently only take one type for the values. (e.g `Dict[str, str]`, `Dict[str, int]`, `Dict[str, float]`, but not,`Dict[str, Any]`). 

The only reason I say this is "broken" is that seeing the train function annotated with

```python
scheduler_kwargs: Dict[str, Any],
optimizer_kwargs: Dict[str, Any]
```
someone would expect to be able to specify these in the `typeo` config, but doing so would break the pipeline.

I'll change it back for the sake of getting this in and open up an issue

deepchatterjeeligo (approved):
LGTM.

PR #49, opened 2023-01-26
Migrate to ml4gw preprocess

EthanMarx:
Migrates whitener and standard scaler to ml4gw versions


libs/trainer/tests/test_wrapper.py
```diff
@@ -43,10 +42,10 @@ def to(self, device):
         return
 
 
-class Preprocessor(Transform):
+class Preprocessor(torch.nn.Module):
```
deepchatterjeeligo, line 45: Wondering why this was a `Transform`

EthanMarx: Don't think it needed to be. If you're curious, the concept has been moved to `ml4gw` and is now called `FittableTransform`

projects/sandbox/train/train.py
```diff
@@ -37,7 +40,7 @@ def load_signals(waveform_dataset: Path, parameter_names: List[str]):
                     values = np.log10(values)
                 data.append(values)
 
-        intrinsic = np.column_stack(data)
+        intrinsic = np.row_stack(data)
```
deepchatterjeeligo, line 43: Was this a bug previously?

EthanMarx: Not a bug: This is a result of moving to `ChannelWiseScaler` in ml4gw (see comment in `preprocessor.py`)

projects/sandbox/train/train.py
```diff
@@ -134,6 +130,8 @@ def main(
     preprocessor.normalizer.fit(intrinsic)
     preprocessor.normalizer.to(device)
 
+    torch.save(preprocessor, outdir / "preprocessor.pt")
```
deepchatterjeeligo, line 133: How do we re-use this?

libs/data/mlpe/data/transforms/preprocessor.py
```diff
@@ -17,32 +15,43 @@ class Preprocessor(torch.nn.Module):
...
 
     def __init__(
         self,
         num_ifos: int,
         sample_rate: float,
-        kernel_length: float,
-        normalizer: Optional["Transform"] = None,
         fduration: Optional[float] = None,
-        highpass: Optional[float] = None,
+        normalizer: torch.nn.Module = torch.nn.Identity(),
```
deepchatterjeeligo, line 36: Why is the normalizer default identity?

projects/sandbox/train/train.py
EthanMarx: I *think* we'll be able to call preprocessor.load_state_dict(). Should double check that

libs/data/mlpe/data/transforms/preprocessor.py
EthanMarx: So that it will just return the parameters themselves if a normalizer isn't passed

deepchatterjeeligo: Minor nitpick. But can we call it something different. Normalizer sounds like it will always normalize. Maybe `scaler`?

EthanMarx: Sure

libs/trainer/pyproject.toml
```diff
@@ -8,16 +8,25 @@ packages = [{include = "mlpe"}]
 
 [tool.poetry.dependencies]
 python = ">=3.8,<3.11"
+torch = {version = "^1.12", source = "torch"}
```
deepchatterjeeligo, line 11: Just a quick check, as a part of this maintenance patch. Shall we remove the minor tick constraint. I noticed some of the projects/libs were installing 1.13 and others were install 1.12. It's not a problem, but I am wondering if we should just do
```suggestion
torch = {version = ">=1.12", source = "torch"}
```

libs/trainer/tests/test_wrapper.py
deepchatterjeeligo: Fittable transform seem like a better option here, with a `build` function etc. no?

deepchatterjeeligo:
This looks good. I'll approve it now. Feel free to merge when you have some of the minor details above done.

deepchatterjeeligo (approved):

projects/sandbox/train/train.py
EthanMarx: Fixed this so that we save the `whitener` and `scaler` individually

libs/trainer/pyproject.toml
EthanMarx: Remind me: what is the exact difference here? I agree should definitely standardize versions

PR #50, opened 2023-02-13
bug fixes to account for ml4gw api changes

EthanMarx:
- Addresses bug in the train project caused by change in `ml4gw` Whitening API 
- Addresses bug due to difference in expected shape of `intrinsic` parameters array in the ml4gw `ChannelWiseScaler` and `RandomWaveformInjection` objects

### DIFF
--- a/libs/data/mlpe/data/transforms/preprocessor.py
+++ b/libs/data/mlpe/data/transforms/preprocessor.py
@@ -49,9 +49,9 @@ class Preprocessor(torch.nn.Module):
         # we need to transpose the parameters tensor so that
         # we normalize along each individual attribute, and not each
         # individual sample
-        parameters.transpose(0, 1)
-        normed = self.normalizer(parameters)
+        parameters = parameters.transpose(1, 0)
+        normed = self.scaler(parameters)
         # transpose back to for consistency with flow.log_prob,
-        normed.transpose(0, 1)
+        normed = normed.transpose(1, 0)
 
         return x, normed
--- a/projects/sandbox/train/train.py
+++ b/projects/sandbox/train/train.py
@@ -124,7 +124,7 @@ def main(
         num_ifos,
         sample_rate,
         fduration,
-        normalizer=standard_scaler,
+        scaler=standard_scaler,
     )
 
     preprocessor.whitener.fit(kernel_length, *background)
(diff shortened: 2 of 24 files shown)


projects/sandbox/train/train.py
```diff
@@ -75,14 +75,17 @@ def main(
     # and validation if valid_frac specified
     background = load_background(background_path, ifos)
 
+    # intrinsic parameters is an array of shape (n_params, n_signals)
     signals, intrinsic = load_signals(waveform_dataset, inference_params)
 
     if valid_frac is not None:
         background, valid_background = split(background, 1 - valid_frac, 1)
```
deepchatterjeeligo, line 82: In line with the previous comment above, let's also add a comment here about the shape of the background

projects/sandbox/train/train.py
```diff
@@ -121,18 +124,23 @@ def main(
         num_ifos,
         sample_rate,
         fduration,
-        normalizer=standard_scaler,
+        scaler=standard_scaler,
     )
 
-    preprocessor.whitener.fit(background)
+    preprocessor.whitener.fit(kernel_length, *background)
```
deepchatterjeeligo, line 130: How has the behavior changed now?

libs/data/mlpe/data/transforms/preprocessor.py
```diff
@@ -49,9 +49,9 @@ def forward(self, strain: torch.Tensor, parameters: torch.Tensor):
         # we need to transpose the parameters tensor so that
         # we normalize along each individual attribute, and not each
         # individual sample
-        parameters.transpose(0, 1)
```
deepchatterjeeligo, line 52: Does any of the tests use this preprocessor? I found this one: https://github.com/ML4GW/PE/blob/main/libs/data/tests/test_dataloader.py#L88-L91 which is defining its own preprocessor. Can this test be changed to use this preprocessor class

deepchatterjeeligo: Also, in general, adding a unittest that shows what are the shapes of the inputs for whitener.fit or scaler.fit will be good

projects/sandbox/train/train.py
EthanMarx: Behavior hasn't changed, just the `ml4gw` API

libs/data/mlpe/data/transforms/preprocessor.py
EthanMarx: Since the preprocessor is essentially a light wrapper of `ml4gw` modules that are already tested, I didnt think we needed any. I think as long as we explain via comments all the odd transposes we have to do we should be good

projects/sandbox/train/train.py
EthanMarx: sounds good

EthanMarx:
@deepchatterjeeligo this should be good to go. Was able to get the training running. I agree the shapes can be confusing, but that all ultimately boils down to using `ml4gw`. Like any library, the API needs to be learned through experience.

PR #51, opened 2023-02-14
first commit of sampling project

deepchatterjeeligo:
Draw samples for every item in test set, and output PP plot


projects/sandbox/pyproject.toml
```diff
@@ -98,3 +100,20 @@ kernel_length = "${base.kernel_length}"
...
 logdir = "${base.logdir}"
+
+[tool.typeo.scripts.sample_testing_set]
+ifos = "${base.ifos}"
+kernel_length = "${base.kernel_length}"
+sample_rate = "${base.sample_rate}"
+fduration = "${base.fduration}"
+trigger_distance = 0.0
+model-state-path = "${base.basedir}/training/weights.pt"
+inference_params = ["frequency", "quality", "hrss", "dec", "psi", "phi"]
```
EthanMarx, line 111: Since `inference_params` is used in the `train` project as well, could we upgrade this to a `base` variable?

EthanMarx: Same with `trigger_distance`

projects/sandbox/sampling/sample.py
```diff
@@ -0,0 +1,167 @@
...
+    preprocessor.scaler = torch.load(scaler_path)
+    return preprocessor
+
+
+def _load_test_data(testing_path: Path, inference_params: List[str]):
+    with h5py.File(testing_path, "r") as f:
+        signals = f["injections"][:]
+        params = []
+        for param in inference_params:
+            values = f["phase" if param == "phi" else param][:]
```
EthanMarx, line 35: There is an ambiguity / bug that we should address (which might explain why the phi variable looked so poorly in the PP plot you showed): in `ml4gw`, `phi` refers to `ra`, not the phase.

projects/sandbox/sampling/sample.py
```diff
@@ -0,0 +1,167 @@
...
+    inference_params: List[str],
+    datadir: Path,
+    logdir: Path,
+    outdir: Path,
+    device: str,
+    num_samples_draw: int,
+    verbose: bool = False,
+):
+    device = device or "cpu"
+    # FIXME: fix phi/phase discrepancy
```
EthanMarx, line 85: Ah I see you ran into a problem related to the above comment as well.

projects/sandbox/sampling/sample.py
```diff
@@ -0,0 +1,167 @@
...
+    priors["phi"] = priors["phase"]
+    del priors["phase"]
+
+    configure_logging(logdir / "sample.log", verbose)
+    num_ifos = len(ifos)
+    num_params = len(inference_params)
+
+    logging.info("Initializing model and setting weights from trained state")
+    model_state = torch.load(model_state_path)
+    flow_obj = architecture((num_params, num_ifos, 2048))
```
EthanMarx, line 96: Can we "un-hard code" this: I think it should be `(kernel_length - f_duration) * sample_rate`

EthanMarx (changes requested):
Some minor tweaks but looks good. Main thing I need to resolve is the phi/phase/ra confusion

deepchatterjeeligo:
Thank you @EthanMarx. Made the change. Yeah, let's create an issue with the parameter naming. We can remove the fixme when resolving that.

EthanMarx (approved):

PR #60, opened 2023-03-05
add toggle for gaussian noise

EthanMarx:
Adds a boolean `gaussian` that if True, will generate gaussian noise from the requested data period.

### DIFF
--- a/projects/sandbox/data_generation/data_generation/generate_background.py
+++ b/projects/sandbox/data_generation/data_generation/generate_background.py
@@ -29,6 +29,7 @@ def main(
     logdir: Path,
     gaussian: bool = False,
     psd_file: Optional[Path] = None,
+    is_psd: bool = True,
     force_generation: bool = False,
     verbose: bool = False,
 ):
@@ -134,6 +135,8 @@ def main(
                 # if user passed a psd file, load it into a FrequencySeries
                 # that ml4gw.spectral.normalize_psd can handle
                 frequencies, psd = np.loadtxt(psd_file, unpack=True)
+                if not is_psd:
+                    psd = psd**2
 
                 # ml4gw expects psd to start at 0 Hz, so lets check for that
                 # TODO: implement logic that prepends 0 to psd
--- a/projects/sandbox/data_generation/data_generation/generate_background.py
+++ b/projects/sandbox/data_generation/data_generation/generate_background.py
@@ -1,15 +1,17 @@
 import logging
 from pathlib import Path
-from typing import List
+from typing import List, Optional
 
 import h5py
 import numpy as np
-from data_generation.utils import gaussian_noise_from_gwpy_timeseries
+from data_generation.utils import noise_from_psd
 from gwdatafind import find_urls
+from gwpy.frequencyseries import FrequencySeries
 from gwpy.segments import DataQualityDict
 from gwpy.timeseries import TimeSeries
 from typeo import scriptify
 
+from ml4gw.spectral import normalize_psd
 from mlpe.logging import configure_logging
 
 
@@ -25,9 +27,10 @@ def main(
     minimum_length: float,
     datadir: Path,
     logdir: Path,
+    gaussian: bool = False,
+    psd_file: Optional[Path] = None,
     force_generation: bool = False,
     verbose: bool = False,
-    gaussian: bool = False,
 ):
     """Generates background data for training BBHnet
 
@@ -37,6 +40,12 @@ def main(
         ifos: which ifos to query data for
         outdir: where to store data
     """
+
+    if psd_file is not None and not gaussian:
+        raise ValueError(
+            "Cannot generate gaussian noise from"
+            " requested PSD when gaussian is False"
+        )
 
     # make logdir dir
     logdir.mkdir(exist_ok=True, parents=True)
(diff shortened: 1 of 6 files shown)


projects/sandbox/data_generation/data_generation/generate_testing_set.py
```diff
@@ -33,6 +37,7 @@ def main(
     min_duration: float = 0,
     waveform_arguments: Optional[Dict] = None,
     parameter_conversion: Optional[Callable] = None,
+    gaussian: bool = False,
```
deepchatterjeeligo, line 40: update docstring.

deepchatterjeeligo (changes requested):
thanks @EthanMarx. Requesting a couple of changes here: but let's a path to the asd file used to generate the noise. Secondly, the noise_from_psd is a relatively small wrapper around lalsimulation. We can reimplement that here instead of adding a large library like pycbc. Finally, please add a unittest with white noise producing a delta-function-like time series, maybe approx matching to height of delta function, or something equivalent.

projects/sandbox/data_generation/data_generation/generate_testing_set.py
EthanMarx: Got it will update doc string. 

I use the requested period of real data to compute the psd, so no need to specify a path.

projects/sandbox/data_generation/data_generation/utils.py
```diff
@@ -4,10 +4,27 @@
 import numpy as np
 import torch
 from gwpy.timeseries import TimeSeries, TimeSeriesDict
+from pycbc.noise import noise_from_psd
```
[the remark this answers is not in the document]
EthanMarx: I completely agree, I was thinking the same thing. I did this out of laziness as `lalsimulation` is really not a friendly api, especially since you don't get descriptive error messages. I was trying to implement that exact function you linked and was running into problems. Can I suggest we open this as an issue for now?

projects/sandbox/data_generation/data_generation/utils.py
```diff
@@ -4,10 +4,27 @@
...
+from pycbc.types import FrequencySeries
 
+from ml4gw.spectral import normalize_psd
 from ml4gw.utils.slicing import slice_kernels
 
 
+def gaussian_noise_from_gwpy_timeseries(
```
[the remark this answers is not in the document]
EthanMarx: Given this function wraps `ml4gw` (which is tested) and `pycbc` (which is presumably tested), do you think this is necessary? When we add our own implementation definitely agree will need a test.

projects/sandbox/data_generation/data_generation/generate_testing_set.py
deepchatterjeeligo: OK, but we can very likely be asked to compare posteriors across different observing runs, and possibly next gen instruments. Besides the changes to achieve that is pretty minimal compared to what you have here.

projects/sandbox/data_generation/data_generation/utils.py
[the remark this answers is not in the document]
deepchatterjeeligo: How about something adapted from the following
```
import lal, lalsimulation
import matplotlib.pyplot as plt

flow = 40.0
duration = 16.0
srate = 16384.0
length = int(duration * srate)
stride = length // 2

epoch = lal.LIGOTimeGPS(0, 0)
rng = lal.gsl_rng("ranlux", 0)

seg = lal.CreateREAL8TimeSeries("STRAIN", epoch, 0.0, 1.0/srate, lal.StrainUnit, length)
psd = lal.CreateREAL8FrequencySeries("LIGO SRD", epoch, 0.0, 1.0/duration, lal.SecondUnit, length//2 + 1)

lalsimulation.SimNoisePSDFromFile(psd, flow, './AplusDesign.txt')
lalsimulation.SimNoise(seg, 0, psd, rng)

times = [seg.deltaT * ii for ii in range(seg.data.length)]
plt.plot(times, seg.data.data)
plt.show()
```
image

I took it from the docs here: https://git.ligo.org/lscsoft/lalsuite/-/blob/master/lalsimulation/lib/LALSimNoise.c#L104-118. For longer lengths, the above function with stride = 1, 2,... until desired length is achieved. The noise curve file is taken from here: https://dcc.ligo.org/LIGO-T2000012/public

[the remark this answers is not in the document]
EthanMarx: Thanks for doing this Deep, will implement and get rid of pycbc.

projects/sandbox/data_generation/data_generation/generate_testing_set.py
EthanMarx: Okay are you thinking of adding an argument like 

```python
psd: Optional[Path] = None
```
If this is set and `gaussian = True`, we use for the psd, otherwise we use the requested data

EthanMarx:
@deepchatterjeeligo Added the requested changes. Let me know what you think.

projects/sandbox/data_generation/data_generation/generate_testing_set.py
deepchatterjeeligo: Yes exactly, a path and a `is_asd` which is a bool, since many of the psd files are actually asds.

projects/sandbox/data_generation/data_generation/generate_background.py
```diff
@@ -88,33 +99,62 @@ def main(
...
             )
 
-            # resample
-            data = data.resample(sample_rate)
+        if gaussian:
+            logging.info(f"Generating gaussian noise from psd for ifo {ifo}")
+
+            # TODO: what should determine this?
+            df = 1 / 8.0
+            if psd_file is not None:
```
deepchatterjeeligo, line 133: Let's add a flag, something like `is_asd` since some of the noise in DCC are asd instead of psd.

projects/sandbox/data_generation/pyproject.toml
```diff
@@ -14,6 +17,7 @@ mldatafind = {git = "https://github.com/ML4GW/mldatafind.git"}
 "mlpe.logging" = {path = "../../../libs/logging", develop = true}
 "mlpe.injection" = {path = "../../../libs/injection", develop = true}
 typeo = {git = "https://github.com/ML4GW/typeo.git"}
+pycbc = "^2.1.0"
```
deepchatterjeeligo, line 20: No longer needed

deepchatterjeeligo (approved):
@EthanMarx I added a small change to the test. Please take a look. Otherwise OK.

deepchatterjeeligo (changes requested):
Sorry, missed one thing `gaussian_noise_from_gwpy_timeseries` is not used. Let's remove that.

deepchatterjeeligo (approved):

PR #61, opened 2023-03-06
Bilby pipe

EthanMarx:
Incorporate analysis of events with bilby


deepchatterjeeligo:
Requesting some high-level changes on this one @EthanMarx.
- [ ] break up the merge request into two
- [ ] The natural split I can think of is the analysis vs the generation. But feel free to do a different break.
- [ ] For the generation part, re-using data generation tools for both the flows vs bilby_pipe workflow would be nice.
- [ ] Regarding the analysis part, the content of the `bilby` project can be moved into the `sample` project since it falls under sampling. The plotting utilities like pp-plots can go into a separate `utils.py` under the `sample` project.
- [ ] Regarding command line tool for sampling, there should be a way to separate the sampling the trained flow vs launching a PE run on the testing set. This could be two different executables, or a single executable with different arguments or subcommands, I leave that up to you.

PR #74, opened 2023-04-04
Coherent dense embedding

EthanMarx:
Adds a linear embedding network that embed coherently


EthanMarx:
@deepchatterjeeligo 

This just adds a `CoherentDenseEmbedding` and a `Resnet`. I didn't change the embedding that we use in this PR, nor the scheduler / optimizer.

PR #76, opened 2023-04-06
typeo updates

EthanMarx:
Cleaned up version of #70.


libs/architectures/mlpe/architectures/__init__.py
```diff
@@ -1 +1,67 @@
...
+
+        return f
+
+    params = inspect.signature(arch).parameters
+    params = list(params.values())[2:]
+    func.__signature__ = inspect.Signature(params)
+    return func
+
+
+# this is a decorator that takes an embedding
```
deepchatterjeeligo, line 26: remove comments

deepchatterjeeligo: In fact let's move the comment as a docstring to the `embeddings`  and `flows` variables below. Add usage for future additions.

libs/architectures/mlpe/architectures/embeddings/dense.py
```diff
@@ -54,18 +54,22 @@ class NChannelDenseEmbedding(nn.Module):
...
-        in_shape,
-        out_shape,
+        shape: Tuple[int, int],
+        context_dim: int,
         **kwargs,
     ) -> None:
         super().__init__()
-        self.N = N
+        n_ifos, in_features = shape
         self.activation = kwargs.get("activation", torch.relu)
```
deepchatterjeeligo, line 63: Would you remove this line @EthanMarx. I probably put it there, I don't know what why. `activation` does not need a special default.

libs/architectures/mlpe/architectures/embeddings/resnet.py
```diff
@@ -302,6 +302,10 @@ def __init__(
         norm_groups: Optional[int] = None,
     ) -> None:
         super().__init__()
+
+        # hack since other embedding require both n_ifos and strain
```
deepchatterjeeligo, line 306: Could you explain this to me

libs/architectures/mlpe/architectures/flows/flow.py
```diff
@@ -73,12 +73,14 @@ def to_device(self, device):
...
-        sample_tensor = torch.zeros((1, self.n_ifos, self.strain_dim))
-        context_dim = self.embedding_net(sample_tensor).shape[-1]
+        dummy_tensor = torch.zeros((1, self.n_ifos, self.strain_dim))
+        context_dim = self.embedding_net(dummy_tensor).shape[-1]
         return context_dim
 
     @property
     def flow(self):
         if self._flow is None:
-            self.build_flow()
```
deepchatterjeeligo, line 83: OK with this change. But I'm curious why this is better choice?

libs/architectures/mlpe/architectures/flows/flow.py
```diff
@@ -73,12 +73,14 @@ def to_device(self, device):
...
+            raise ValueError(
+                "Flow is not built. Call build_flow before accessing flow"
```
deepchatterjeeligo, line 84:
```suggestion
                "Flow is not built. Call build_flow() before accessing flow"
```

libs/architectures/mlpe/architectures/flows/flow.py
deepchatterjeeligo: This original code acted like a singleton. If the flow was built, it would give the flow object, else build the flow and returned it.

libs/trainer/mlpe/trainer/optimizers.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(
+        self,
+        parameters: Iterable,
+        lr: float = 1e-3,
+        weight_decay: float = 0.0,
+    ):
+        super().__init__(parameters, lr=lr, weight_decay=weight_decay)
+
+
+optimizers = {"adam": _wrap_optimizer(Adam)}
```
deepchatterjeeligo, line 34: same comments at the wrappers around flow and embeddings. Add variable docstring, and describe usage, and future addition directives.

libs/trainer/mlpe/trainer/optimizers.py
```diff
@@ -0,0 +1,34 @@
+import inspect
+from typing import Iterable
+
+import torch
+
+
+def _wrap_optimizer(optimizer):
```
deepchatterjeeligo, line 7: Can we try to get this double-level wrapper inside __init__.py, and call it something like `_core_wrapper`, and reuse it for both optimizer and scheduler. In one case, the first arg is the model parameters, in the other it is the optimizer, but the wrapper is agnostic to that (at least looks like so)

projects/sandbox/pyproject.toml
```diff
@@ -162,6 +185,13 @@ device = "cuda"
...
-commands.coupling = "${base.coupling}"
-commands.maf = "${base.maf}"
+# flow subcommand
+[tool.typeo.scripts.sample-with-flow.flow]
+maf = "${base.maf}"
+coupling = "${base.coupling}"
+
+# embedding subcommand
+[tool.typeo.scripts.sample-with-flow.embedding]
+dense = "${base.dense}"
```
deepchatterjeeligo, line 195: I am a little confused here. Which one is being used? Asking since the `train:maf` under `steps` above does not specify the embedding.

deepchatterjeeligo (changes requested):
Hi @EthanMarx, requesting some changes and some questions above, but good job overall. thanks.

libs/architectures/mlpe/architectures/embeddings/resnet.py
EthanMarx: Yeah so the embeddings require you pass `n_ifos` and `strain_dim`, but the `resnet` we've built determines this on its own. 

So, to make instantiation of the embedding in the `trainer` lib the same for all of the embeddings, I include this hack.

libs/architectures/mlpe/architectures/flows/flow.py
EthanMarx: This is a relic of when I had a setup where build_flow() took the `embedding` as an argument. 

I can change it back.

libs/trainer/mlpe/trainer/optimizers.py
EthanMarx: Yup thats a good point. Let me try to generalize it.

projects/sandbox/pyproject.toml
EthanMarx: Yeah good question. So with this new `typeo` update, the `pinto.steps` table is deprecated. When alec writes a fix for that the syntax will look something like the following:

```toml
[[tool.pinto.steps]]
project = train
script = train
architecture = maf
embedding = resnet
optimizer = adam
scheduler = onecycler
 ...
```

So for now, the scripts need to be run by hand. The syntax for that is also slightly different:

```
pinto run train --typeo ../pyproject.toml script=train flow=maf embedding=resnet optimizer=adam scheduler=onecycle
```

Note that the order of the subcommands must match the order they are specified in the `@scriptify` wrapper

Issue #82, opened 2023-05-03
Refactor/move sample_with_flow project content into exisiting projects


deepchatterjeeligo:
Right now `sample` project draws samples from the _large_ testing set and creates a PP plot. We would also like to run the flow on the same injections on which we ran bilby. To do this,

- [ ] Move the bilby injection data generation into the `data_generation` project, and dump the data in both formats: one that bilby wants for running dynesty, and one that the `sample` project wants. This data_generation script should have the ability to write the bilby config file pointing to the data format that bilby wants, and have a `--submit` argument to call `bilby_pipe`.  
- [ ] With the above change, the existing executable in the `sample` project can be run on the injections that bilby is run on. Add an additional `--bilby-result-dir` argument (or something similar) to the existing `sample` project. If this argument is supplied, the code will plot overlapping corner plots using both the bilby results and the samples from the flow.

These change will go into the aps-dev branch.

deepchatterjeeligo:
@EthanMarx would you like to handle the first task? I can take care of the second.

EthanMarx:
Sounds like a plan

PR #83, opened 2023-05-08
Unify the sampling and sample-with-flows projects

deepchatterjeeligo:
Addresses second part of #82


deepchatterjeeligo:
@EthanMarx anything else on this?

PR #90, opened 2023-06-01
Use ml4gw sine gaussian

EthanMarx:
Use torch implementation of SG for better sampling of parameter space


libs/injection/mlpe/injection/injection.py
```diff
@@ -1,81 +1,94 @@
...
+    )
+    cross = np.zeros(n_freqs, dtype=complex)
+    plus = np.zeros(n_freqs, dtype=complex)
 
-        # center so that coalescence time is middle sample
-        dt = waveform_duration / 2
-        polarizations = np.roll(polarizations, int(dt * sample_rate), axis=-1)
-        signals[i] = polarizations
+    cross[: len(hcross.data.data)] = hcross.data.data
+    plus[: len(hplus.data.data)] = hplus.data.data
```
deepchatterjeeligo, line 92: why is this assigned relative to the start index of  `plus` or `cross`?

libs/injection/mlpe/injection/injection.py
```diff
@@ -1,81 +1,94 @@
...
+
+        cross.append(hcross)
+        plus.append(hplus)
+
+    cross = np.stack(cross)
+    plus = np.stack(plus)
+    return cross, plus
+
+
+def sine_gaussian_frequency(
```
deepchatterjeeligo, line 49: rename to `generate_frequency_domain_sine_gaussian` similar to its time domain counterpart.

projects/sandbox/sampling/run.sh
```diff
@@ -0,0 +1,2 @@
+pinto run sample-with-flow --typeo ../pyproject.toml script=sample-with-flow flow=maf embedding=resnet
```
deepchatterjeeligo, line 1: This is a good idea. Do you want to add your runner script for the train project as well

projects/sandbox/train/train.py
```diff
@@ -29,30 +47,29 @@ def load_background(background_path: Path, ifos):
 
 def load_signals(waveform_dataset: Path, parameter_names: List[str]):
     """
-    Load in signals and intrinsic parameters.
-    If no intrinsic parameters are requested, return None.
+    Load in validation signals and parameters
```
deepchatterjeeligo, line 50: Let's provide more context here for later that we are loading validation data so that we generate signals using lalsimulation/lalinference.

deepchatterjeeligo:
This is a big change, several files have been touched. Overall it looks good to me. There are a few comments I have left above, mostly queries and some semantics.

Overall training using ml4gw and validating against lalsimulation/lalinference is a good idea.

libs/injection/mlpe/injection/injection.py
EthanMarx: Colm sent me this code at some point. I will look into the details of all of this.

libs/injection/mlpe/injection/injection.py
EthanMarx: Since it's only really meant to be used with bilby, will do `bilby_frequency_domain_sine_gaussian` to clarify that

PR #94, opened 2023-08-02
Bilby pipe overhaul

EthanMarx:
Overhauls bilby data generation and analysis

### DIFF
--- a/projects/sandbox/analyze_with_bilby/analyze_with_bilby/deploy.py
+++ b/projects/sandbox/analyze_with_bilby/analyze_with_bilby/deploy.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from typing import List, Optional
 
 import bilby
+import h5py
 import numpy as np
 from mldatafind.segments import query_segments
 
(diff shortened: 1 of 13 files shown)


projects/sandbox/analyze_with_bilby/analyze_with_bilby/deploy.py
```diff
@@ -0,0 +1,125 @@
...
+        should_transfer_files="YES",
+        when_to_transfer_output="ON_EXIT_OR_EVICT",
+        stream_error=True,
+        stream_output=True,
+        transfer_output_files=outdir,
+        checkpoint_exit_code=77,
+    )
+    dag_id = condor.submit(subfile)
+    logging.info(f"Launching bilby analyses with dag id {dag_id}")
+    condor.watch(dag_id, condordir, held=True)
```
deepchatterjeeligo, line 122: won't this be a blocking call?

deepchatterjeeligo (commented):
This looks good @EthanMarx in terms of the implementation side. My only one comment was about the condor_watch_q command which is a blocking call. We could avoid it since bilby outputs the dlogz to the `*out` files.

PR #96, opened 2023-11-07
reimplement flows in pyro

deepchatterjeeligo:


alecgunny:
Just took a quick look at the code, and out of curiosity is there a reason to not have the `NormalizingFlow` class inherit from `torch.nn.Module`, that way you can inherit all of the `parameters` and `state_dict` functionality?

deepchatterjeeligo:
> is there a reason to not have the NormalizingFlow class inherit from torch.nn.Module

Nope. That should have been done right from the start.

PR #99, opened 2024-03-06
Lightning cli

EthanMarx:
This PR does a bit of restructuring / generalizing with the following high level goals:

## Main goals:
1. Support use of the `LightningCLI` to get rid of `typeo` dependency
2. Generalize our `DataModule` to support both "on the fly" and from disk waveform generation
3. Begin generalizing `DataModule` to be agnostic to waveform type (e.g CBC / sine-Gaussian)

## Additional notable features:
1. Use local whitening in place of a fixed whitener
2. Allow data-loading from multiple background files (i.e multiple segments)

### `BaseDataset` 

`BaseDataset` handles most of the data-loading and preprocessing logic. The idea is to have as subclasses a `WaveformGeneratorDataset` (implemented in this PR) that generates waveforms on the fly and a `FromDiskDataset` (not yet implemented) that will load waveforms from disk. The key methods that these two subclasses need to implement are:

- `fit_scaler`: Defines how to fit the standard scaler. 
    - `WaveformGeneratorDataset`: this amounts to sampling some number of parameters from the prior and using those. 
    - `FromDiskDataset`: the pre-sampled parameters will be loaded from disk and used
- `get_val_waveforms`: Defines how to generate / load validation waveforms. 
    -  `WaveformGeneratorDataset`: Some specified number of waveforms will be generated. Note that these will be split across the number of devices for distributed validation, and each device will have a unique set of waveforms. Note also that these waveforms are generated once up front so that each epoch uses the same exact validation set. 
    - `FromDiskDataset`: Via some parameter like e.g. `valid_frac`, a portion of the waveforms saved on disk will be used for validation. There should also be  logic to split these across devices.
 - `sample_waveforms`: Defines how waveforms are sampled 
     - `WaveformGeneratorDataset`: Parameters are sampled and waveforms generated on the fly
    - `FromDiskDataset`: Waveforms are sampled from pre-computed set in memory. In the future could support loading form multiple waveform files on disk to increase quantity of waveforms available.

    
## Architecture re-structure
- Introduces `FlowArchitecture` class that is a `PyroModule` ( i.e. a `torch.nn.Module`). `FlowArchitecture` takes an `Embedding` as an argument such that embeddings and flows are fully decoupled. `jsonargparse` (the workhorse behind `LightningCLI`) makes it really easy to configure thins like this. For an example, see the example `config.yaml` in this PR. 

- The neural-network architecture used is now fully decoupled from the `pl.LightningModule`. We have done something similar in `aframe`. So now the `LightningModule` looks something like 

```python 
class PEModel(pl.LightningModule):
    def __init__(
        self,
        arch: FlowArchitecture,
        patience: Optional[int] = None,
        save_top_k_models: int = 10
        ....
    ) -> None:
        super().__init__()
        self.model = arch

    def forward(self, parameters, strain) -> Tensor:
        return -self.model.log_prob(parameters, context=strain)

    def training_step(self, batch, _):
        strain, parameters = batch
        loss = self(parameters, strain).mean()
        self.log(
            "train_loss",
            loss,
            on_step=True,
            on_epoch=True,
            prog_bar=True,
            sync_dist=False,
            logger=True,
        )
        return loss
       ....
```

Now, the `LightningModule` is fully agnostic to the exact architecture it is training. As long as the architecture implements the requirements of `FlowArchitecture` it will be good to go. 

## TODOs
Right now the `WaveformGenerator` class is hardcoded to use a `SineGaussian` model. I think the best way to generalize this to arbitrary waveform models is to take @deepchatterjeeligo's `FrequencyDomainWaveformGenerator` object and move it to `ml4gw`.


projects/sandbox/train/train/data/base.py
```diff
@@ -0,0 +1,516 @@
...
+        fnames = list(test_dir.glob("*.hdf5"))
+        return fnames
+
+    def train_val_split(self) -> Sequence[str]:
+        """
+        Split background files into training and validation sets
+        based on the requested duration of the validation set
+        """
+        fnames = sorted(self.train_val_fnames)
+        durations = [int(fname.stem.split("-")[-1]) for fname in fnames]
```
deepchatterjeeligo, line 187: This is making some assumptions about the filenames that did not used to be the case before. Use a regex pattern here to be more robust and explicit. Also, mention in a comment how what is the standard format, for example, gwosc hdf5 format.

projects/sandbox/train/train/data/base.py
```diff
@@ -0,0 +1,516 @@
...
+        valid_fnames = []
+        valid_duration = 0
+        while valid_duration < self.hparams.min_valid_duration:
```
deepchatterjeeligo, line 190: This will throw an error if there is only one background hdf5 file that matches the glob, but is long enough. Can we allow for that?

projects/sandbox/train/train/data/base.py
```diff
@@ -0,0 +1,516 @@
...
+    # ================================================ #
+    # Utilities for initial data loading and preparation
+    # ================================================ #
+    def setup(self, stage: str) -> None:
+        world_size, rank = self.get_world_size_and_rank()
+        self._logger = self.get_logger(world_size, rank)
+
+        # infer sample rate directly from background data
+        with h5py.File(self.train_fnames[0], "r") as f:
+            sample_rate = 1 / f[self.hparams.ifos[0]].attrs["dx"]
```
deepchatterjeeligo, line 227: This "dx" key is also a breaking change, and makes assumptions about the hdf5 dataset. This comment is tied to the one above.

projects/sandbox/train/train/data/generator.py
```diff
@@ -0,0 +1,163 @@
...
+        transformed = self.transform(fit_params)
+        transformed = [v for k, v in transformed.items()]
+        parameters = torch.vstack(transformed)
+
+        scaler.fit(parameters)
+        return scaler
+
+    def get_waveforms(self, num_waveforms: int):
+        cross, plus, parameters = self.sample_waveforms(
+            num_waveforms, device="cpu"
```
deepchatterjeeligo, line 88: Why are we hardcoding the device to cpu here? This will have to be followed by a transfer, which is unnecessary if the data is directly generated on the GPU. Can we not use `self.datamodule.device`?

projects/sandbox/train/train/data/base.py
```diff
@@ -0,0 +1,516 @@
...
+
+        return torch.utils.data.DataLoader(
+            dataset,
+            pin_memory=False,
+            batch_size=batch_size,
+            num_workers=self.num_workers,
+        )
+
+    def train_dataloader(self):
+        dataset = Hdf5TimeSeriesDataset(
```
deepchatterjeeligo, line 495: I'm seeing the source code of the Hdf5 dataset. This does not preserve existing behavior of pushing the entire background to the GPU. From what I understand this will load background on the CPU and do a transfer to the GPU. Is that correct?

projects/sandbox/train/train/model.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def get_logger(self):
+        logger_name = "PEModel"
+        return logging.getLogger(logger_name)
+
+    def on_fit_start(self):
+        datamodule = self.trainer.datamodule
+        for item in datamodule.__dict__.values():
+            if isinstance(item, torch.nn.Module):
+                item.to(self.device)
```
deepchatterjeeligo, line 65: This confirms my comment above using the HDF5 dataset class. Using this trick, we are not really generating the waveforms on the GPU, but transferring it after generating it on the CPU.

projects/sandbox/train/train/model.py
```diff
@@ -0,0 +1,209 @@
...
+            self.hparams.num_samples_draw, context=strain
+        )
+        descaled = self.trainer.datamodule.scale(samples, reverse=True)
+        result = self.cast_as_bilby_result(
+            descaled.cpu().numpy(),
+            parameters.cpu().numpy()[0],
+        )
+        self.test_results.append(result)
+
+        if batch_idx % 100 == 0 and self.num_plotted < self.num_corner:
```
deepchatterjeeligo, line 160: Can we make this `100` a configuration. I had been meaning to do this, but left it hardcoded thus far.

projects/sandbox/train/train/model.py
```diff
@@ -0,0 +1,209 @@
...
+            skymap_filename = self.outdir / f"{self.num_plotted}_mollview.png"
+            corner_filename = self.outdir / f"{self.num_plotted}_corner.png"
```
deepchatterjeeligo, line 162: Let's also dump the result file as json or pickle for the cases where we have the mollview and corner plots.

projects/sandbox/train/train/waveforms.py
```diff
@@ -0,0 +1,112 @@
...
+                A callable that returns cross and plus polarizations
+                given a set of parameters.
+            duration:
+                The duration of the waveform in seconds
+            sample_rate:
+                Sample rate of the waveform in Hz
+            f_min:
+                The minimum frequency of the waveform in Hz
+            f_max:
+                The maximum frequency of the waveform in Hz
```
deepchatterjeeligo, line 33: add docstring for `padding`

deepchatterjeeligo (changes requested):
Hi @EthanMarx this is a good refactor. I like most aspects of it, but I think as it stands now, it is losing some of the features of the InMemoryDatasets from ml4gw. We should preserve that along with this new HDF5 dataset addition.

projects/sandbox/train/train/data/base.py
EthanMarx: Yeah I can add a regex for it: The format is `{prefix}-{t0}-{duration}.hdf5`.

deepchatterjeeligo: I was trying this out with some gwosc background, and now thinking about this further, this also needs the ifo in front. Like `H-H1_GWOSC_O3b_4KHZ_R1-1262243840-4096.hdf5` ; `L-L1_GWOSC_O3b_4KHZ_R1-1262243840-4096.hdf5`. Same under /dev/shm/kafka: `H-H1_O3ReplayMDC_llhoft-1394823647-1.gwf`

deepchatterjeeligo: What I am saying is we should move to something like the online streaming frames instead of using something custom.

projects/sandbox/train/train/model.py
EthanMarx: Waveforms should still be generated on GPU

projects/sandbox/train/train/model.py
EthanMarx: Since our test set is generated randomly I think we can get rid of this right?

projects/train/train/data/datasets/base.py
```diff
@@ -0,0 +1,455 @@
...
+        return strain, parameters
+
+    # ================================================ #
+    # Dataloaders used by lightning
+    # ================================================ #
+
+    def train_dataloader(self):
+        # if we only have one training file
+        # load it into memory and use InMemoryDataset
+        if len(self.train_fnames) == 1:
```
deepchatterjeeligo, line 348: Please add a self.log("Using InMemoryDataset/HDF5Dataset...") statement under each conditional.

projects/train/train/data/datasets/base.py
```diff
@@ -0,0 +1,455 @@
...
+        waveform_dataloader = torch.utils.data.DataLoader(
+            waveform_dataset,
+            batch_size=self.val_batch_size,
+            shuffle=False,
+            pin_memory=False,
+        )
+
+        # build background dataloader
+        val_background = self.val_background[0][:, rank:]
+
```
deepchatterjeeligo, line 397: Add self.log statement here about the InMemoryDataset usage (like above)

projects/train/train/data/datasets/base.py
```diff
@@ -0,0 +1,455 @@
...
+            batch_size=self.val_batch_size,
+            batches_per_epoch=len(waveform_dataloader),
+            coincident=False,
+            shuffle=False,
+        )
+
+        background_dataloader = torch.utils.data.DataLoader(
+            background_dataset, pin_memory=False
+        )
+        return ZippedDataset(
```
deepchatterjeeligo, line 410: What is the reason behind returning both dataloaders for validation and testing?

projects/train/train/data/waveforms/generator/cbc.py
```diff
@@ -0,0 +1,131 @@
...
+                The maximum frequency of the waveform in Hz
+            padding:
+                The amount of padding to add to the right
+                of the waveform in seconds
+            waveform_arguments:
+                A dictionary of fixed arguments to pass to the waveform model
+        """
+        super().__init__(*args, **kwargs)
+        waveform_arguments = waveform_arguments or {}
+        self.waveform = partial(approximant, **waveform_arguments)
```
deepchatterjeeligo, line 50: What keyword arguments are you passing via `waveform_arguments`?

projects/train/train/data/waveforms/generator/cbc.py
```diff
@@ -0,0 +1,131 @@
+import math
+from functools import partial
+from typing import Any, Callable, Dict
+
+import torch
+from train.data.waveforms.generator.generator import WaveformGenerator
+
+
+class CBCGenerator(WaveformGenerator):
```
deepchatterjeeligo, line 9: Let's add FrequencyDomain in the name, since this will work when the callable returns h(f)

deepchatterjeeligo (changes requested):
@EthanMarx I like this. I have left some comments. In terms of design this is great. Once you have completed a run with this, let's get this in.

projects/train/train/cli.py
```diff
@@ -0,0 +1,80 @@
+from lightning.pytorch.cli import LightningCLI
+from train.data.datasets.base import AmplfiDataset
+from train.models.base import AmplfiModel
+
+
+class AmplifiCLI(LightningCLI):
```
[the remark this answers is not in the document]
EthanMarx: Good catch

projects/train/train/data/datasets/flow.py
```diff
@@ -0,0 +1,38 @@
+import torch
+from train.data.datasets.base import AmplfiDataset
+
+
+class FlowDataset(AmplfiDataset):
+    """
+    Lightning DataModule for training normalizing flow networks
```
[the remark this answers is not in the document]
EthanMarx: Yup will do that. I plan on adding documentation to each method in this whole object.

projects/train/train/data/waveforms/generator/cbc.py
EthanMarx: For CBC, I think the only one is `f_ref`. The idea here is to fix certain arguments that aren't randomly sampled

projects/train/train/data/datasets/base.py
EthanMarx: So for validation and testing I wanted to make sure that things are deterministic between each epoch so that our validation score is directly comparable. So, I generate the validation and testing waveforms once up front before training begins.

Then, the waveforms and background are loaded together and injected in the `on_after_batch_transfer` function

projects/train/train/losses.py
```diff
@@ -0,0 +1,46 @@
...
+
+        std_x = torch.sqrt(x.var(dim=0) + 0.0001)
+        std_y = torch.sqrt(y.var(dim=0) + 0.0001)
+        std_loss = (
+            torch.mean(F.relu(1 - std_x)) / 2
+            + torch.mean(F.relu(1 - std_y)) / 2
+        )
+
+        x = (x - x.mean(dim=0)) / x.std(dim=0)
+        y = (y - y.mean(dim=0)) / y.std(dim=0)
```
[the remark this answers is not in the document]
EthanMarx: To clarify, we don't need to divide by the `std`?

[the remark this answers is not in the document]
deepchatterjeeligo: right we do not need to divide by std, since we want the full covariance

projects/train/train/architectures/flows/maf.py
```diff
@@ -0,0 +1,57 @@
+import torch
+import torch.distributions as dist
+from pyro.distributions.conditional import ConditionalComposeTransformModule
+from pyro.distributions.transforms import ConditionalAffineAutoregressive
+from pyro.nn import ConditionalAutoRegressiveNN
+from train.architectures.flows import FlowArchitecture
+
+
+class MaskedAutoRegressiveFlow(FlowArchitecture):
```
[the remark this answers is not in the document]
deepchatterjeeligo: For IAF: https://arxiv.org/pdf/1606.04934 for MAF: https://arxiv.org/abs/1705.07057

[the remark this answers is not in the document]
deepchatterjeeligo: Both are based on this paper: https://proceedings.mlr.press/v37/germain15.html which first reports the use of masked linear layers for auto-regressive transforms.

projects/train/apptainer.def
```diff
@@ -0,0 +1,37 @@
...
+# are installed editably, so we'll opt for using
+# the full poetry machinery without using a
+# virtual environment. Note that this can potentially
+# cause issues if the poetry install makes changes to
+# the virtualenv package in the base environment, so
+# we'll hard pin it to the value we need.
+# TODO: there's got to be a better solution here
+python -m pip install poetry
+cd /opt/amplfi/projects/train
+poetry export -o requirements.txt --without-hashes \
```
deepchatterjeeligo, line 24: Why not simply
```
python -m pip install --no-deps -r <(poetry export --without-hashes)
pip install --no-deps .
```

projects/train/apptainer.def
```diff
@@ -0,0 +1,37 @@
...
+Bootstrap: docker
+From: python:3.10.12-slim-bullseye
+Stage: final
+
+%files from build
+/opt/amplfi /opt/amplfi
+
+%post
+cd /opt/amplfi/projects/train
+pip install -r requirements.txt --no-deps --no-cache-dir
```
deepchatterjeeligo, line 36: this part is not needed, see comment above

projects/train/configs/flow/cbc.yaml
```diff
@@ -0,0 +1,82 @@
...
+seed_everything: 101588
+trainer:
+  accelerator: auto
+  strategy: auto
+  devices: auto
+  profiler: "simple"
+  logger:
+    - class_path: lightning.pytorch.loggers.WandbLogger
+      init_args:
+        save_dir: "/home/ethan.marx/amplfi/runs/first-lightning-run/train/"
```
deepchatterjeeligo, line 11: Is there a way to get this from an environment variable. Otherwise this will be a diff everytime based on who creates an MR

projects/train/apptainer.def
EthanMarx: Yup, think that one liner would work too let me try it out

projects/train/configs/flow/cbc.yaml
EthanMarx: Yeah good point there definitely is let me dig into it.

projects/train/train/data/waveforms/generator/cbc.py
deepchatterjeeligo: OK great. Let's add that under the docstring i.e. `A dictionary of fixed arguments to pass to the waveform model. E.g. ``f_ref`` which stays constant for the run.`

projects/train/configs/flow/cbc.yaml
deepchatterjeeligo: We can use the link_arguments feature here with

```
compute_fn=lambda x: os.getenv(x) if x.startswith("$") else x
```
or something equivalent to attach the paths. Need to be careful of
```
    save_dir: "$AMPLFI_SAVE_DIR"
```
vs `${AMPLFI_SAVE_DIR}` but that's a minor detail.

projects/train/train/architectures/losses.py
```diff
@@ -0,0 +1,41 @@
+"""Module storing loss functions"""
+import torch
+import torch.nn.functional as F
+from torch import nn
+
+
+# implemented from https://github.com/violatingcp/codec
+class VICRegLoss(nn.Module):
+    def forward(self, x, y, wt_repr=1.0, wt_cov=1.0, wt_std=5.0):
```
deepchatterjeeligo, line 9: Let's keep the default to be `1` for the weights.

deepchatterjeeligo (approved):
Looks good.

EthanMarx:
Great gonna merge - most of the tests are obsolete now. We can add those workflows back in slowly overtime

PR #100, opened 2024-05-14
Update `ml4gw` submodule commit

EthanMarx:
For some reason the `ml4gw` submodule was pointing to a stale commit. This points to the `v0.4.2` tag.

For now, this also removes the stale unit testing workflow.


deepchatterjeeligo:
Do we need a submodule any longer? Specially if we use a tag

deepchatterjeeligo (changes requested):
Let's either use the tag: https://github.com/ML4GW/amplfi/blob/main/projects/train/pyproject.toml#L11 . Or, since my sense is we will likely be using ml4gw `dev`, let's point to that branch instead.

EthanMarx:
I think we should submodule, in case we need to iterate faster than `ml4gw` deployments

EthanMarx:
@deepchatterjeeligo Alright we're pointing at dev

deepchatterjeeligo:
sorry for the delay. thanks.

deepchatterjeeligo:
This is not what I was asking, but fine for now.

deepchatterjeeligo (approved):

PR #104, opened 2024-06-20
Incorporate `law` and `mldatafind`

EthanMarx:
Incorporates recent `mldatafind` overhaul so we can now generate training and testing strain data via `law`. 

Updates the README with minimal instructions on getting started.


deepchatterjeeligo:
can you separate out the sorting of the imports in a separate PR @EthanMarx

EthanMarx:
yeah sure

EthanMarx:
@deepchatterjeeligo Should be good for review now that pre commit is fixed

amplfi/law/pyproject.toml
```diff
@@ -0,0 +1,18 @@
...
+version = "0.1.0"
+description = "Law tasks for running amplfi analyses"
+authors = ["Ethan Jacob Marx <ethan.marx@ligo.org>"]
+packages = [{include = "amplfi"}]
+
+
+[tool.poetry.dependencies]
+python = ">=3.9,<3.12"
+law = {git = "https://github.com/riga/law.git", branch = "master"}
+mldatafind = {path = "../../mldatafind/", develop = true}
```
deepchatterjeeligo, line 12: Do we have this submodule?

amplfi/law/amplfi/law/base.py
```diff
@@ -0,0 +1,31 @@
+import os
+
+from amplfi.law.paths import paths
+
+
+class AmplfiDataTaskMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
deepchatterjeeligo, line 8: What is the super class here?

amplfi/law/amplfi/law/base.py
```diff
@@ -0,0 +1,31 @@
...
+        self.image = paths().container_root / "data.sif"
+
+    def sandbox_env(self, env):
+        # get `mldatafind.law.DataTask` environment variables
```
deepchatterjeeligo, line 12: Can you provide a use case example here i.e. what `env`

amplfi/law/pyproject.toml
deepchatterjeeligo: I see its added

deepchatterjeeligo (changes requested):
I think this mostly looks fine. If you can remove the stray notebook, and explain some of the questions I have.

amplfi/law/amplfi/law/base.py
EthanMarx: Yeah this is a `law` hook for mapping local environment variables into the container that the tasks are run inside. So for `mldatafind` these are things like credentials, segment database urls, etc.

amplfi/law/pyproject.toml
EthanMarx: yup

amplfi/law/amplfi/law/base.py
EthanMarx: This is meant to be a "Mixin" for `law` `SandboxTasks`, specifically for the use case of appending environment variables. You can see it's use e.g. [here](https://github.com/ML4GW/amplfi/blob/19cdf5ba5646f7724f2ba0976ea1596e90f4a318/amplfi/law/amplfi/law/tasks.py#L24)

For reasons that are hard to explain over github, it didn't make sense to subclass the `mldatafind` tasks and override the `sandbox_env` due to additional boiler plate that would induce

amplfi/law/amplfi/law/Untitled-1.ipynb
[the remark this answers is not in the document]
EthanMarx: done

amplfi/law/amplfi/law/base.py
deepchatterjeeligo: OK. But then add comments to which method in the law/mldatafind source code is being invoked when you call `super()`.

amplfi/law/amplfi/law/base.py
deepchatterjeeligo: Let's add that as a docstring, otherwise its unclear.

EthanMarx: done

amplfi/law/amplfi/law/base.py
EthanMarx: done

EthanMarx:
@deepchatterjeeligo Just added some more documenation. Let me know if things are still unclear

deepchatterjeeligo (approved):

PR #107, opened 2024-06-23
Update similarity config with `outdir`

EthanMarx:


EthanMarx:
@deepchatterjeeligo Just a config file fix here so gonna take the liberty to merge

PR #111, opened 2024-08-06
add maf; reinstate tests

deepchatterjeeligo:


amplfi/architectures/pyproject.toml
```diff
@@ -2,14 +2,17 @@
...
-authors = ["Ethan Jacob Marx <ethan.marx@ligo.org>"]
+authors = ["Ethan Jacob Marx <ethan.marx@ligo.org>", "Deep Chatterjee <deep.chatterjee@ligo.org>"]
 packages = [{include = "amplfi"}]
 
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
-pyro-ppl = "^1.9.0"
+ml4gw = {path = "../../ml4gw", develop = true}
+pyro-ppl = "*"
```
EthanMarx, line 12: Is there a reason we're unconstraining `pyro`? I imagine there's _some_ version where our code will break.

EthanMarx (commented):
Looks good. Thanks for adding the testing infra back in. Just a question about version constraints for pyro (and pytest as well)

amplfi/architectures/pyproject.toml
deepchatterjeeligo: Usually we should constrain a version when a feature breaks, and mention the issue from the corresponding repo. So far I have not found our pyro features to break.

deepchatterjeeligo: But at the same time, I don't feel very strongly.

deepchatterjeeligo: done

EthanMarx (approved):

PR #113, opened 2024-08-06
add searched area calculation to testing stage

deepchatterjeeligo:
* minor fixes in coordinate transformation in healpy arrays
* add a new method to set the healpix array and searched area for Result objects


deepchatterjeeligo:
does this look good @EthanMarx

EthanMarx (approved):

PR #115, opened 2024-08-09
Simplification of law sandboxing

EthanMarx:
Simplifies the law Sandboxing by moving environment variable setting to the `Sandbox` (as opposed to the `Task`). This allows users to more easily subclass the `mldatafind` `DataSandbox` w/o having to fiddle with the tasks.


deepchatterjeeligo:
OK this somewhat makes sense. I'll go ahead and merge this.

PR #117, opened 2024-08-12
Add checkpointing and improve flow logging

EthanMarx:


EthanMarx:
@ravioli1369 once merged, adding `checkpoint` under the model init_args should work with the `test` subcommand

projects/train/configs/flow/cbc.yaml
```diff
@@ -20,6 +20,8 @@ model:
   class_path: train.models.flow.FlowModel
   init_args:
     outdir: ${oc.env:AMPLFI_OUTDIR}
+    # uncomment below when running `trainer.test` to load in a checkpoint
+    # checkpoint:
```
deepchatterjeeligo, line 24: This looks good. Is it possible to keep something like Null or None value instead of uncommenting?

EthanMarx: Ah yeah good call I can make it `null`

EthanMarx: done

PR #120, opened 2024-08-13
load checkpoint

aggar104:


projects/train/train/cli/sim_runner.slurm
EthanMarx, line 1: Same as above: shouldnt be in the repo

projects/train/configs/similarity/cbc.yaml
```diff
@@ -20,7 +20,6 @@ model:
   class_path: train.models.SimilarityModel
   init_args:
     outdir: ${oc.env:AMPLFI_OUTDIR}
-    checkpoint: null
```
EthanMarx, line 23: Is there a reason you removed this? In principle it shouldn't break anything, but I may be wrong.

projects/train/configs/similarity/cbc.yaml
```diff
@@ -11,7 +11,7 @@ trainer:
         save_dir: ${oc.env:AMPLFI_OUTDIR}/train
         project: amplfi
         # uncomment and set to name your run
-        # name: 
+        name: emeddings_1
```
EthanMarx, line 14: Reset this to `# name`

projects/train/configs/flow/cbc.yaml
```diff
@@ -57,9 +55,9 @@ data:
     psd_length: 10
     fftlength: 2
     batches_per_epoch: 200
-    batch_size: 512
+    batch_size: 800
     ifos: ["H1", "L1"]
-    min_valid_duration: 10000.0
```
EthanMarx, line 62: I think keeping a higher (10000) minimum validation duration now makes sense

projects/train/configs/flow/cbc.yaml
```diff
@@ -20,19 +20,17 @@ model:
...
         hidden_features: 150
         num_transforms: 80
         num_blocks: 6
-        context_dim: 7
+        context_dim: 8
         # uncomment below to load 
         # in pre-trained embedding weights
-        # embedding_weights: "path/to/embedding/weights"
-        # freeze_embedding: false
+        embedding_weights: "/projects/bcse/saggarwal/binds/checkpoints/last.ckpt"
```
EthanMarx, line 32: Change these back to the commented versions you see above

projects/train/configs/flow/cbc.yaml
```diff
@@ -20,19 +20,17 @@ model:
   class_path: train.models.flow.FlowModel
   init_args:
     outdir: ${oc.env:AMPLFI_OUTDIR}
-    # add path below when running `trainer.test` to load in a checkpoint
-    checkpoint: null
```
EthanMarx, line 24: Was having `checkpoint = null` breaking things?

amplfi/architectures/amplfi/architectures/flows/base.py
```diff
@@ -28,8 +29,25 @@ def __init__(
         else:
             self.embedding_context = nullcontext
 
-        if embedding_weights is not None:
-            self.embedding_net.load_state_dict(torch.load(embedding_weights))
+        def load_weights(weights: Path):
```
EthanMarx, line 32: Great stuff. We still wan't the ability to not pass the `embedding_weights`. So a conditional like
```python
if embedding_weights is not None:
    self.embedding_net.load_state_dict(embedding_dict)
```

Also, I don't think we should be loading in the standard scaler weights here. Were you running into a problem?

aggar104: There was no problem without scaler weights. Can we ignore them, or do they need to be loaded at a later stage?

projects/train/configs/flow/cbc.yaml
aggar104: I think I was using an older config file that didn't have this command. It didn't give an error with this.

amplfi/architectures/amplfi/architectures/flows/base.py
EthanMarx: If you're starting a flow training, they'll just be recomputed again (and should be essentially the same). 

We can ignore them here.

projects/train/configs/flow/cbc.yaml
EthanMarx: Got it. If you rebase your branch onto the latest changes from main, that should fix it. 

I'm also noticing you are making this PR from your main branch. In the future you should make a feature branch and submit the PR from there.

PR #122, opened 2024-08-14
add feature to oversample projection parameters

deepchatterjeeligo:
After this patch, the waveform generator will have the ability to oversample dec, phi, psi by a factor >= 1, so that the effective batch size is batch_size * oversampling_factor.

### DIFF
--- a/projects/train/configs/flow/cbc.yaml
+++ b/projects/train/configs/flow/cbc.yaml
@@ -83,5 +83,6 @@ data:
         f_min: 20
         f_max: 1024
         padding: 1
+        oversampling_factor: 3
         waveform_arguments:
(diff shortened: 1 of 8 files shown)


deepchatterjeeligo:
#125 to be considered first.

ravioli1369:
I feel the effective batch_size = `batch_size` * `oversampling_factor` should be made a bit clearer in the config, maybe `oversampling_factor` can be a param just right below `batch_size`, with a comment that explains it, or the `batch_size` can stay the same, and internally `batch_size//oversampling_factor` waveforms would be generated, then oversampled (this also should include a comment next to `batch_size`)

PR #123, opened 2024-08-16
Remove ml4gw as submodule

wbenoit26:


EthanMarx:
I think this is good. Will enlist @deepchatterjeeligo as well just to confirm

.gitmodules
```diff
@@ -1,9 +1,6 @@
 [submodule "hermes"]
 	path = hermes
 	url = https://github.com/ML4GW/hermes
```
deepchatterjeeligo, line 3: can you remove this too, likely a remnant.

amplfi/architectures/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 1.8.3 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
```
deepchatterjeeligo, line 1: can you regenerate this using 1.8.3, followed by `poetry lock --no-update`.

amplfi/architectures/pyproject.toml
```diff
@@ -8,7 +8,7 @@ packages = [{include = "amplfi"}]
 
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12"
-ml4gw = {path = "../../ml4gw", develop = true}
+ml4gw = "^0.5"
```
deepchatterjeeligo, line 11: Use ">=0.5", we can restrict if the major tick breaks things

projects/train/pyproject.toml
```diff
@@ -8,7 +8,7 @@ license = "MIT"
 [tool.poetry.dependencies]
 python = ">=3.9,<3.12" #^3.8,<3.12"
 lightning = "^2.2.1"
-"ml4gw" = {path = "../../ml4gw", develop = true}
+ml4gw = "^0.5"
```
deepchatterjeeligo, line 11: ">=0.5"

EthanMarx: I think maybe `>=0.4` ? The issue we ran into was some conflicts with `aframe` libs, so in case other projects have this as a dep would be nice to have some wiggle room

wbenoit26: I've done `>=0.4`, let me know if that works @deepchatterjeeligo

deepchatterjeeligo (approved):
LGTM

PR #125, opened 2024-08-22
device consistency

deepchatterjeeligo:
refactoring device consistency from #122. This moves the project, psd estimator, and waveform generator to the right device.

### DIFF
--- a/projects/train/train/data/waveforms/generator/cbc.py
+++ b/projects/train/train/data/waveforms/generator/cbc.py
@@ -48,7 +48,7 @@ class FrequencyDomainCBCGenerator(WaveformGenerator):
                 e.g. `f_ref` for CBC waveforms
         """
         super().__init__(*args, **kwargs)
-        waveform_arguments = waveform_arguments or {}
+        self.waveform_arguments = waveform_arguments or {}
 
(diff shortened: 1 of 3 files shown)


projects/train/train/data/waveforms/generator/cbc.py
```diff
@@ -103,8 +103,13 @@ def time_domain_strain(self, **parameters):
                 the right of the waveform in seconds
         """
 
-        # TODO: support time domain waveforms
-        hc, hp = self.waveform(self.frequencies[self.freq_mask], **parameters)
+        device = parameters["chirp_mass"].device
+        freqs = torch.clone(self.frequencies).to(device)
```
ravioli1369, line 107: What's the reason for cloning the frequency grid here? Does passing it into the approximant change it?

deepchatterjeeligo: The `self.frequency` can be on the cpu. So I clone it and transfer to the gpu for use in this function, without impacting the global attribute.

deepchatterjeeligo:
anything else on this?

ravioli1369:
nope

PR #133, opened 2024-09-11
Make `FrequencyDomainCBCGenerator` robust to choices of `f_min`, `f_max`, and `padding`

EthanMarx:
@ravioli1369 ran into an issue where the waveforms being generated were not sufficiently long enough. The underlying issue was that the number of samples in the time domain data after ffting the waveform is a function of `f_min` and `f_max`.

Another issue was the ringdown of the ffted waveform was on the left side of the waveform.

This PR does a couple of things to address these:

1. Adds `RINGDOWN_DURATION` class variable set to `0.5` seconds that will roll the waveform by that amount so that the coalescence and ringdown are properly connected. This means that by default, the waveform coalesence time is `RINGDOWN_DURATION` from the right edge of the kernel.
2. The `padding` variable  now specifies _additional_ zero padding on top of the `0.5` seconds of ringdown to add to the waveform
3. A `left_padding` variable is calculated to ensure that the waveform is long enough to fulfill the requested duration

### DIFF
--- a/projects/train/train/data/waveforms/generator/cbc.py
+++ b/projects/train/train/data/waveforms/generator/cbc.py
@@ -68,14 +68,17 @@ class FrequencyDomainCBCGenerator(WaveformGenerator):
     @property
     def right_pad_size(self):
         """
-        User defined value that controls where the coalescence time
-        lies relative to the right edge
+        Size of additional right padding in samples
         """
         return math.ceil(self.padding * self.sample_rate)
 
     @property
     def left_pad_size(self):
-        """ """
+        """
+        Size of left padding required to ensure
+        the waveform is sufficiently long to slice
+        according to the user requested `duration`
+        """
         # calculate the size of the time domain
         # waveform after ffting
         freq_dim = self.freq_mask.sum()
--- a/projects/train/train/data/waveforms/generator/cbc.py
+++ b/projects/train/train/data/waveforms/generator/cbc.py
@@ -9,11 +9,36 @@ from train.data.waveforms.generator.generator import WaveformGenerator
 
 class FrequencyDomainCBCGenerator(WaveformGenerator):
     """
-    A torch module for generating CBC waveforms on the fly using
-    ml4gw waveform models.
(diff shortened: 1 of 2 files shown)


projects/train/train/data/waveforms/generator/cbc.py
```diff
@@ -9,12 +9,37 @@
...
+            padding on top of that.
+        waveform_arguments:
+            A dictionary of fixed arguments to pass to the waveform model,
+            e.g. `f_ref` for CBC waveforms
     """
 
+    # duration of ringdown - will ensure
+    # coalescence time is at least this
+    # far from the right edge
+    RINGDOWN_DURATION = 0.5
```
deepchatterjeeligo, line 41: Let's make this configurable with a default of 0.5s

EthanMarx: The main reason I decided not to make it configurable in the `__init__` was to protect users from fiddling with it and breaking things.

deepchatterjeeligo: yeah but the ringdown is different based on the priors, similar to f_max. You can add a note in the docs about it if you like to.

EthanMarx: done

ravioli1369:
the ringdown duration should also be defined in the `cbc.yaml` file as well right? since it's configurable

EthanMarx:
> the ringdown duration should also be defined in the `cbc.yaml` file as well right? since it's configurable

Yeah I think specifying it in the config makes sense - let me do that

EthanMarx:
@ravioli1369 I set `ringdown_duration = 0.5` and `padding = 0.0` as defaults in the config, which should place the coalescence 0.5 seconds from the right edge of the window. 

Let me know if any issues come up with this

ravioli1369:
The rest looks good, I tried generating a waveform as before and the shifts look right. Also started a training run to see how this compares to the model I trained with the old waveform generator.

PR #140, opened 2024-10-07
Checkpoint

aggar104:
Add Checkpoints


amplfi/architectures/amplfi/architectures/flows/base.py
```diff
@@ -28,8 +29,25 @@ def __init__(
         else:
             self.embedding_context = nullcontext
 
-        if embedding_weights is not None:
```
EthanMarx, line 31: @aggar104 We should still have this check for `embedding_weights is not None`. We still want to support use cases where we train the embedding flow end to end from scratch. 

This is why the tests were failing - our unit tests don't pass embedding weights to the model. 

Also, could you move the `load_weights` definition out of the `__init__` and to a class method?

EthanMarx:
@aggar104 I think this has been addressed in a separate PR, so going to close it for now.

PR #141, opened 2024-10-07
Automate searched area plot

EthanMarx:


projects/train/train/models/flow.py
```diff
@@ -161,9 +162,25 @@ def test_step(self, batch, _):
...
+        searched_areas = []
+        for result in self.test_results:
+            searched_area = result.calculate_searched_area()
+            searched_areas.append(searched_area)
+        searched_areas = np.sort(searched_areas)
+        counts = np.arange(1, len(searched_areas) + 1) / len(searched_areas)
+
+        plt.figure(figsize=(10, 6))
+        plt.step(searched_areas, counts, where="post")
+        plt.xlabel("Searched Area (deg^2)")
```
deepchatterjeeligo, line 183: Generally, a log scaling on the x-axis is able to show the plot better, but it may not always work. Shall we plot two versions of the plot? One with the log scale and one with linear scale?

deepchatterjeeligo (commented):
Looks good. Only a minor comment about the axis scaling.

projects/train/train/models/flow.py
EthanMarx: I think you're right. I actually haven't fully tested this , working on it now. Let me do that and I'll post the plot in here

EthanMarx: Okay got it going - Yes xaxis log scale looks much better just pushed

deepchatterjeeligo (approved):

PR #142, opened 2024-10-08
Restructure into single module

EthanMarx:
Moves everything under the `amplfi` package. There are now `train` , `data` and `tune` submodules. amplfi is now pip installable.

The `apptainer.def` file contains all the necessary requirements for running data querying training and tuning. 

The `law` data querying currently requires using `apptainer`, but training and tuning do not.


deepchatterjeeligo (approved):
The diff is significant, but overall this looks great. So let's give it a shot, and we will adjust anything that was missed subsequently.

PR #149, opened 2024-10-11
docs touch up

deepchatterjeeligo:


docs/installation.md
```diff
@@ -52,15 +52,15 @@ Then you can either pull the container from the remote github repository
...
+        $ apptainer pull $AMPLFI_CONTAINER_ROOT/amplfi.sif docker://ghcr.io/ml4gw/amplfi/data:main
 
     Supported python versions: 3.9-3.12.
 
   .. tab:: docker
 
     .. code-block:: console
 
-      $ docker pull ghcr.io/ml4gw/amplfi:main $AMPLFI_CONTAINER_ROOT/amplfi.sif 
+      $ docker pull ghcr.io/ml4gw/amplfi/data:main
```
EthanMarx, line 63: The `amplfi` container should be the one stop shop for both datagen and training. I should remove the `data` and `train ` container releases from the repo

EthanMarx:
@deepchatterjeeligo Thanks for this. Also, there is a typo referencing Aframe in the `containers.md` docs [here](https://github.com/ML4GW/amplfi/blob/main/docs/containers.md?plain=1#L14) would be great if you could fix that up as well.

docs/installation.md
deepchatterjeeligo: done

PR #152, opened 2024-10-14
Install `amplfi` editably so that code can be bind mounted

EthanMarx:


deepchatterjeeligo:
This is a very helpful feature with apptainer. Do we have any docs related to how to bind the git repo? If not, that will be a good addition, specially for those starting out.

EthanMarx:
Definitely (also sorry for taking liberty of merging some of these PRs last night, was trying to get tuning going on nautilus). 

I can add that here: `https://amplfi.readthedocs.io/en/stable/containers.html`

PR #154, opened 2024-10-14
Change install path of `mamba` env in container

EthanMarx:
1. Changes install path of `mamba` env in container due to conflict when mounting code on nautilius
2. Adds documentation on mounting local code into containers at runtime


EthanMarx:
@deepchatterjeeligo gonna merge this in to get the container build updated so I can test out tuning, but let me know if anything looks off.

PR #156, opened 2024-10-14
Callback updates

EthanMarx:
Update callbacks to be compatible with `s3`


EthanMarx:
@deepchatterjeeligo this PR just made our callbacks compatible with `s3` so tuning would run on nautilius

deepchatterjeeligo (approved):
Looks good

PR #164, opened 2024-10-24
Mdc Dataset

EthanMarx:


EthanMarx:
@deepchatterjeeligo take a look. 

(note I also added infra for using `asd` info. I have our dataloaders / augmentors automatically pass this along to the embedding, and let the embedding decide whether or not it wants to use it)

amplfi/train/data/datasets/testing/debug-1.ipynb
deepchatterjeeligo, line 1: Let's add `*ipynb` to gitignore

amplfi/train/data/datasets/testing/mdc.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # scale parameters
+        parameters = self.scale(parameters)
+
+        # calculate asds
+        freqs = torch.fft.rfftfreq(X.shape[-1], d=1 / self.hparams.sample_rate)
+        num_freqs = len(freqs)
+        psds = torch.nn.functional.interpolate(
+            psds, size=(num_freqs,), mode="linear"
+        )
```
deepchatterjeeligo, line 116: what is this for?

deepchatterjeeligo (commented):
Looks good overall @EthanMarx. Have you made a diagnostic plots for the strain after whitening to ensure that things look good?

amplfi/train/data/datasets/testing/mdc.py
EthanMarx: this makes sure the psds are the same length of the data so they can be concatenated before passing to the embedding. (This is done under the hood in the `Whitener`)

EthanMarx:
Yeah I made qscans and strain plots and things looked fine. The signals were placed at the proper location.

amplfi/train/data/datasets/testing/mdc.py
deepchatterjeeligo: Let's put that down as a one-line comment?

deepchatterjeeligo (approved):

PR #174, opened 2024-11-13
Adds optional `jitter` parameter

EthanMarx:
Adds `jitter` float parameter that specifies, in seconds, how much the coalescence time can vary from the default location 
For CBC waveforms this default is `ringdown_duration` + `padding` seconds from the right edge of the kernel. 

So if `jitter = 0.1` the waveform coalescence can be `+/- 0.1` from the default location.


deepchatterjeeligo (approved):
Looks good

PR #181, opened 2024-11-22
add docs for HPO on SLURM

deepchatterjeeligo:


deepchatterjeeligo:
@EthanMarx heard that you are in transit. I'll pull this in. And in fact it might be easier to read the rendered docs.

PR #188, opened 2024-12-13
Parameterize `nside` and plots skymaps using count `density`

EthanMarx:
Exposes `nside` as a paramter to the model and calculates skymaps as a count density of samples


deepchatterjeeligo (approved):
Good change

PR #193, opened 2025-01-08
compute distance moments from samples

deepchatterjeeligo:


EthanMarx:
Looks good. Is the idea that once we get the skymap stuff working we'll just plug in `result.dist_mu` etc.

deepchatterjeeligo:
> Looks good. Is the idea that once we get the skymap stuff working we'll just plug in `result.dist_mu` etc.

Yeah we can attach the distance parameters to the result object and dump it as fits.

PR #198, opened 2025-01-13
compute distance ansatz per pixel

deepchatterjeeligo:
* Restrict to fixed nside
* Compute moments from samples in each pixel
* Estimate ansatz parameters for pixel that have a good number of pixels


EthanMarx (approved):
It looks good. A couple of suggestions but I'll leave them up to you. 

1. Maybe consider moving the bayestar ansatz definitions to a different file 
2. If possible, would be good to move the distance ansatz calculation portion under `get_sky_projection` to its own function.

deepchatterjeeligo:
@EthanMarx took care of your changes.

PR #205, opened 2025-02-07
Migrate to ml4gw `TimeDomainCBCGenerator` for better waveform conditioning

EthanMarx:


EthanMarx:
Relevant run using this branch https://wandb.ai/ml4gw/amplfi/runs/ysdhdby7?nw=nwuserethanmarx

amplfi/train/data/waveforms/generator/cbc.py
```diff
@@ -1,153 +1,41 @@
...
-        waveform_arguments:
-            A dictionary of fixed arguments to pass to the waveform model,
-            e.g. `f_ref` for CBC waveforms
-    """
-
+class CBCGenerator(WaveformGenerator):
     def __init__(
         self,
         *args,
         approximant: Callable,
```
deepchatterjeeligo, line 13: Let's add a good docstring to this, giving an example of the callable (possibly `ml4gw.waveforms.IMRPhenomD()` and similar?). Does the waveform instance be in the correct device at the time of calling the init?

amplfi/train/data/waveforms/sampler.py
```diff
@@ -34,6 +34,7 @@ class WaveformSampler(torch.nn.Module):
 
     def __init__(
         self,
+        *args,
         fduration: float,
         kernel_length: float,
```
deepchatterjeeligo, line 39: Let's update the docstring for this.

amplfi/train/priors.py
```diff
@@ -27,12 +27,103 @@ def sg_prior() -> ParameterSampler:
     )
 
 
+def precessing_to_lalsimulation_parameters(
+    parameters: dict[str, torch.Tensor]
+) -> dict[str, torch.Tensor]:
+    """
+    Convert precessing spin parameters to lalsimulation parameters
```
deepchatterjeeligo, line 34: Can we add a test here to test to check batched input?

deepchatterjeeligo (changes requested):
@EthanMarx Looks good overall. Requesting a few small changes in terms of docstrings and adding a unittest for the precessing parameter conversions.

amplfi/train/data/waveforms/generator/cbc.py
EthanMarx: Docstring added. The whole `amplfi.train.waveforms.generator.WaveformGenerator` is a torch module that should get sent to the proper device, and thus any sub-torch modules like the waveform instance should get sent there too.

amplfi/train/priors.py
EthanMarx: For which part? The `bilby_to_lalsim_spins` and `chirp_mass_and_mass_ratio_to_components` have tests on the ml4gw side already that test for batched input

deepchatterjeeligo: I see. (Commented without checking that). All good then.

PR #206, opened 2025-02-07
Functions for filtering out values being outside the prior in flow.py

asasli:
1. `_get_log_prior_dict` : Constructs a dictionary for the priors (also adds boundaries for the missing ones).
2. `filter_descaled_parameters` : Filters the descaled posteriors based on the imposed prior and prints the number of discarded values (both total and per parameter).
3. Updated `test_step` to call `filter_descaled_parameters`.
The final **descaled** posteriors will be within the prior boundaries.

### DIFF
--- a/amplfi/train/models/flow.py
+++ b/amplfi/train/models/flow.py
@@ -46,6 +46,8 @@ class FlowModel(AmplfiModel):
 
         # save our hyperparameters
         self.save_hyperparameters(ignore=["arch"])
+        # create a dictionary of prior objects for each parameter
+        self.log_prior_dict = self._get_log_prior_dict()
 
     def forward(self, context, parameters) -> Tensor:
         return -self.model.log_prob(parameters, context=context)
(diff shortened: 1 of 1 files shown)


amplfi/train/models/flow.py
```diff
@@ -124,6 +126,66 @@ def on_test_epoch_start(self):
...
+                log_prior_dict[param] = sf.parameters[param]
+            else:
+                # others are seperately
+                # waveform_sampler."name"
+                # 'dec': Cosine(),
+                # 'psi': Uniform(low: 0.0, high: 3.140000104904175),
+                # 'phi': Uniform(low: -3.140000104904175, high: 3.140000104904175)
+                log_prior_dict[param] = getattr(self.trainer.datamodule.waveform_sampler, param)
+        # add low and high boundaries (for some parameters these values are missing)
+        log_prior_dict['distance'].low = 100
```
EthanMarx, line 157: @asasli Im wondering if, instead of using the prior ranges, we can simply call `prior.prob(sample)`, and discard the elements that have 0 probability. That way, this functionality is robust to changes in the prior ranges, and we don't have to add these prior specific `low` and `high` attributes. Do you think that would work?

Alternatively, we could add `low` and `high` attributes to each of the prior classes.

amplfi/train/models/flow.py
```diff
@@ -124,6 +126,66 @@ def on_test_epoch_start(self):
...
+        valid_idxs = torch.ones(descaled.shape[0], dtype=torch.bool, device=descaled.device)
+        num_discarded = 0
+        for idx, param in enumerate(self.inference_params):
+            prior = self.log_prior_dict[param]
+            low = prior.low
+            high = prior.high
+            valid_idxs &= (descaled[:, idx] >= low) & (descaled[:, idx] <= high)
+            discarded_count = (~valid_idxs).sum().item()
+            num_discarded += discarded_count
+            print(f"Discarded samples[{param}]: {discarded_count}")
```
EthanMarx, line 183: Instead of print, you should be able to use `self._logger.info` to use the model logger object

amplfi/train/models/flow.py
```diff
@@ -133,6 +195,7 @@ def test_step(self, batch, _):
         )
         descaled = self.trainer.datamodule.scale(samples, reverse=True)
         parameters = self.trainer.datamodule.scale(parameters, reverse=True)
+        descaled = self.filter_descaled_parameters(descaled) # filter out samples outside of prior boundaries
```
EthanMarx, line 198: In general, we typically add comments above the relevant code line in this repo. Of course, nothing wrong with adding them inline, but good to keep style consistent

amplfi/train/models/flow.py
```diff
@@ -124,6 +124,69 @@ def on_test_epoch_start(self):
         self.test_results: list[Result] = []
         self.idx = 0
 
+    def _get_log_prior_dict(self, trainer=None):
+        """
+        Get the log prior for each parameter in the model.
+        """
+        if trainer is not None:
+            self.trainer = trainer
```
deepchatterjeeligo, line 132: This step should not be necessary. (I know we talked about this; mentioning here for my own sake).

deepchatterjeeligo (commented):
I would be interested to see whether setting `validate_args=False` in the priors resolves most of the issue you are seeing.

EthanMarx:
@asasli going to close this, since got the relevant code in #211

PR #207, opened 2025-02-11
Adds `ParameterTestingDataset` for performing injections from a parameter set

EthanMarx:
Many things that need to be cleaned up but wanted to open for visibility


amplfi/train/testing.py
```diff
@@ -27,7 +27,8 @@ def get_sky_projection(ra, dec, dist, nside=32, min_samples_per_pix=15):
     """
     theta = np.pi / 2 - dec
     # mask out non physical samples;
-    mask = (ra > 0) * (ra < 2 * np.pi)
+    # mask = (ra > 0) * (ra < 2 * np.pi)
+    mask = (ra > -np.pi) * (ra < np.pi)
```
deepchatterjeeligo, line 31: Let's remove this from version control, and stick to the 0 to 2*pi convention (same as ligo.skymap). If you have a trained model that you are working with, you can apply this patch locally.

poetry.lock
deepchatterjeeligo, line 1: this MR needs to be rebased.

deepchatterjeeligo (commented):
Two small points, otherwise looks good.

deepchatterjeeligo:
Let's merge #211 first

EthanMarx:
@deepchatterjeeligo should be good for one last look over

EthanMarx:
@deepchatterjeeligo one more quick look over here

deepchatterjeeligo (approved):
Looks good.

PR #208, opened 2025-02-14
`StrainVisualization` plotting callback

EthanMarx:
Adds a lightning callback for plotting strain and asd data from the testing step. This was extremely useful for debugging some of the issues we were seeing with the MDC. 

I also added an optional `test_outdir` argument to the `pl_module` that defaults to `train_outdir / test`. This is to remove the need to constantly change the `$AMPLFI_OUTDIR` environment variable. 

Along these lines, the testing directory now gets really cluttered with plots and needs some organization. I was thinking of making either
- A subdirectory for each event (e.g. `{test_outdir}/event_0/` that contains each of the made plots
- A subdirectory for each type of plot (e.g. `{test_outdir}/skymaps/event_0.png` 

Each has its pros and cons and shouldn't matter too much in the end, but wanted your opinion @deepchatterjeeligo

### DIFF
--- a/amplfi/train/callbacks.py
+++ b/amplfi/train/callbacks.py
@@ -185,18 +185,23 @@ class StrainVisualization(pl.Callback):
 
         # unpack batch
         strain, asds, parameters = batch
+        strain, asds = strain[0], asds[0]
 
         # steal some attributes needed from datamodule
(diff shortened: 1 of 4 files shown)


deepchatterjeeligo (approved):
Very useful.

PR #209, opened 2025-02-17
Migrate to `astral`'s `ruff` and `uv` project management tools

EthanMarx:
Ended up being pretty straightforward. 

There are some linting / formatting rules that `ruff` doesn't yet support. But, not having them didn't seem to affect any of our code. 

The reason for a lot of the file changes in this PR is the inclusion of [this](https://docs.astral.sh/ruff/rules/missing-trailing-comma/) missing trailing comma rule, which is supposed to reduce the sizes of diffs when changing things like function signatures.


deepchatterjeeligo (approved):
Let's give this a shot.

PR #210, opened 2025-02-17
update definition files to use uv

deepchatterjeeligo:
Install deps from uv.lock when building image


EthanMarx (approved):

deepchatterjeeligo:
BTW @EthanMarx since you asked, I am not seeing a significant improvement in terms of building the container (takes about 5 mins) since exporting the deps is not the time consuming step. In fact the biggest contribution is the writing SIF file step.

EthanMarx:
yeah that makes sense.

PR #214, opened 2025-02-19
Quality of life improvements when using s3

EthanMarx:
- Adds `max_num_workers` argument to dataset
- Demotes downloading file logging statements to debug
- Suppresses botocore logging when downloading from s3


deepchatterjeeligo (approved):
Looks good.

PR #215, opened 2025-02-21
Adds testing dataset for analyzing real events without injections

EthanMarx:
- Adds a testing dataset that, given a list or file with gpstimes, can find the corresponding raw strain segments from the background files in the background directory for analysis - useful for analyzing real events.
- When testing, the scaler is no longe refit


EthanMarx:
@deepchatterjeeligo wanted to bump this

deepchatterjeeligo (approved):
Sorry about the delay. Thanks @EthanMarx

PR #220, opened 2025-03-06
Organize testing output directories and Add Condorized ligo-skymap-from-samples

EthanMarx:
- Each event during testing now has its own output directory
- Adds `LigoSkymap` law task for condorizing ligo-skymap-from-samples on a set of amplfi posteriors


deepchatterjeeligo:
Good feature. Merge when ready.

PR #222, opened 2025-03-10
Add gradient tracking callback

EthanMarx:
Adding callback to monitor gradient magnitudes to help debug issues like this one https://wandb.ai/ml4gw/amplfi/runs/ag20vpx9?nw=nwuserethanmarx where there was a huge spike in training and validatoin loss.

Likely due to a bad batch, but would be good to be robust to this by adding in gradient clipping


deepchatterjeeligo (approved):
Looks great. Do you want to update the config to add this callback?

EthanMarx:
I think will keep it out by default, since most people wont really need / use it

PR #224, opened 2025-03-27
Migrate to Zuko and update configs

EthanMarx:
- Migrates NF architecture building to zuko
- Updates configuration files to use most recent best hyper parameters, and to use Pv2 by default


EthanMarx:
@wbenoit26 give this a look. Im going to look over myself as well.

PR #225, opened 2025-03-27
Refactor `Result` object

EthanMarx:
Refactors the skymap calculation code outside of the `Result` object so that it can be used on its own without having to instantiate a Result. 

Moves a couple of submodules that didn't really belong under the `train` module into a new `utils` module. 

Adds `searched_volume` calculation


EthanMarx:
@ravioli1369 This likely changes a few of your PRs related to the online code in aframe, but I think it should make things cleaner. Let me know what you think.

amplfi/train/models/flow.py
```diff
@@ -121,7 +138,7 @@ def test_step(self, batch, batch_idx) -> bilby.result.Result:
         # plot corner and skymap
         skymap_filename = test_outdir / "mollview.png"
         corner_filename = test_outdir / "corner.png"
-        fits_filename = test_outdir / "amplfi.skymap.fits"
+        # fits_filename = test_outdir / "amplfi.skymap.fits"
```
wbenoit26, line 141: Should this just be deleted?

amplfi/train/models/flow.py
```diff
@@ -151,11 +167,10 @@ def predict_step(self, batch, batch_idx, _):
         test_outdir = self.test_outdir / f"event_{batch_idx}"
         test_outdir.mkdir(parents=True, exist_ok=True)
 
-        result.calculate_skymap(self.nside, self.min_samples_per_pix)
         skymap_filename = test_outdir / f"{gpstime.item()}_mollview.png"
         corner_filename = test_outdir / f"{gpstime.item()}_corner.png"
-        fits_filename = test_outdir / f"{gpstime.item()}.fits"
-        result_filename = test_outdir / f"{gpstime.item()}_result.hdf5"
+        # fits_filename = test_outdir / f"{gpstime.item()}.fits"
```
wbenoit26, line 172: Same here?

wbenoit26 (approved):
I think this all makes sense to me, just a couple places with extraneous comments.

For my own understanding: what's happening during the "testing" phase? Does the pipeline just switch into analyzing some test dataset after training?

EthanMarx:
Yeah, when you call `trainer.test` (in our case we have a cli wrapper `amplfi-flow-cli test --config config.yaml`)

lightning will use the `test_dataloader` defined by whichever data module instance you've passed in the `config.yaml`, and run `test_step` defined by your lightning data module on the batches returned by `test_dataloader`

We've taken advantage of this to define multiple different `LightningDataModules` with different behavior for loading data to accomodate all of our use cases (e.g. loading in pre-made test injections from an hdf5 file, or generating injections based on a list of parameters). 

So, you can just adjust the `data.class_path` argument in the `config.yaml` to point to whichever data loader corresponds to your use case.

amplfi/train/models/flow.py
EthanMarx: Lets keep this b/c at some point I'll re add ths in

wbenoit26:
I see, that's pretty handy

PR #228, opened 2025-04-17
`mldatafind` update

EthanMarx:
Update `mldatafind` version to include a new feature that allows using lal's resampling algorithm (which is more like `scipy.signal.resample` than `gwpy.timeseries.TimeSeries.resample`) and is actually the default for bilby


wbenoit26:
Should also update the GitHub actions versions here

.github/workflows/pre-commit.yaml
```diff
@@ -10,10 +10,10 @@ jobs:
...
+    - uses: actions/checkout@v4
       with:
         submodules: recursive
-    - uses: actions/setup-python@v2
+    - uses: actions/setup-python@v5
       with:
-        python-version: "3.9"
-    - uses: pre-commit/action@v2.0.3
+        python-version: "3.10"
+    - uses: pre-commit/action@v3.0.01
```
wbenoit26, line 19: Typo - should be `v3.0.1`

wbenoit26 (approved):

PR #229, opened 2025-04-21
Save raw data when using `StrainVisualization` callback

EthanMarx:


EthanMarx:
> Looks good! Any reason for preferring `.txt` files?

Nope - should we prefer .npy?

wbenoit26:
No idea. Probably doesn't matter for how it's being used here.

PR #230, opened 2025-04-28
Generalize priors `conversion_function`

EthanMarx:
Generalize priors `conversion_function` to pass through spin params so they can be used for inference


EthanMarx:
Yeah thats a good point to keep in mind, they get soaked up into the **kwargs of the waveforms

PR #233, opened 2025-05-10
Support timeslides

EthanMarx:
Support analyzing timeshifted data during testing step for `RawStrainTestingDataset`

### DIFF
--- a/amplfi/train/data/datasets/base.py
+++ b/amplfi/train/data/datasets/base.py
@@ -474,7 +474,6 @@ class AmplfiDataset(pl.LightningDataModule):
 
         # build background dataloader
         val_background = self.val_background[0][:, rank:]
-
         background_dataset = InMemoryDataset(
             val_background,
             kernel_size=int(self.hparams.sample_rate * self.sample_length),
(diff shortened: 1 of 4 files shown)


amplfi/train/data/datasets/base.py
```diff
@@ -545,7 +544,10 @@ def inject(self, *args, **kwargs):
         raise NotImplementedError
 
     def background_from_gpstimes(
-        self, gpstimes: np.ndarray, fnames: List[Path]
+        self,
+        gpstimes: np.ndarray,
+        fnames: List[Path],
+        timeslides: Optional[np.ndarray] = None,
```
deepchatterjeeligo, line 550: should we raise a RunTimeError checking that the length of `gpstimes` and `timeslides` is the same.

deepchatterjeeligo (approved):
Just left a minor comment, but looks good otherwise. Approving.

amplfi/train/data/datasets/base.py
EthanMarx: good idea

PR #235, opened 2025-05-12
Specify `ParameterSampler` prior via the config

EthanMarx:
This PR makes the `ParameterSampler` (i.e. our training prior) able to be specified directly in the config. This is important for keeping track of which training prior was used, which previously was specified directly in the code. This could lead to discrepancies if the defaults in the code were changed. 

To make sure the config isn't getting to long, I've specified an additional `priors.yaml` that is then referenced from within the main `cbc.yaml` config. 

This also allows the prior config to be saved in WandB


deepchatterjeeligo (approved):
I like this refactor. Looks good.

PR #245, opened 2025-06-03
Improve validation background dataset construction

EthanMarx:
- Validation background segments are now drawn from multiple files instead of just one
- Use `concurrent.futures` instead of `multiprocessing` when aggregating crossmatch stats due to SIGTERMs being sent to processes


deepchatterjeeligo (approved):
Looks good.

PR #247, opened 2025-06-09
Add ability to reweight to different prior

EthanMarx:
- Adds ability to reweight to a different prior (specified by a bilby prior file) in post processing
- To facilitate the above, refactors some things outside of the `waveform_sampler` into the `BaseDataset`
- Fixes some callbacks to work with the `predict_step`
- `ParameterSampler` is renamed to `AmplfiPrior` as I think its a more apt name

### DIFF
--- a/amplfi/utils/result.py
+++ b/amplfi/utils/result.py
@@ -10,6 +10,7 @@ import matplotlib.pyplot as plt
 import numpy as np
 from ligo.skymap.postprocess.crossmatch import crossmatch, CrossmatchResult
 from copy import copy
+import pandas as pd
 
 
 class AmplfiResult(bilby.result.Result):
@@ -115,15 +116,29 @@ class AmplfiResult(bilby.result.Result):
         return fig
 
     def reweight_to_prior(
-        self, target_prior: bilby.core.prior.PriorDict
+        self,
+        target_prior: bilby.core.prior.PriorDict,
+        max_rejection_steps: int = 1,
     ) -> "AmplfiResult":
-        reweighted = copy(self)
+        reweighted_result = copy(self)
         keys = list(target_prior.keys())
         samples_dict = {key: self.posterior[key].values for key in keys}
         target_probs = target_prior.ln_prob(samples_dict, axis=0)
         ln_weights = target_probs - self.posterior["log_prior"].values
         weights = np.exp(ln_weights)
-        reweighted.posterior = bilby.core.result.rejection_sample(
-            reweighted.posterior, weights
-        )
-        return reweighted
+
+        num_samples = len(self.posterior)
+        reweighted_posterior = pd.DataFrame(columns=self.posterior.columns)
+        for _ in range(max_rejection_steps):
+            if len(reweighted_posterior) >= num_samples:
+                break
+
+            samples = bilby.core.result.rejection_sample(
+                self.posterior, weights
+            )
+            reweighted_posterior = pd.concat(
+                [reweighted_posterior, samples], ignore_index=True
+            )
+
+        reweighted_result.posterior = reweighted_posterior[:num_samples]
+        return reweighted_result


EthanMarx:
@deepchatterjeeligo when you get a chance, could you take a look at this?

amplfi/train/callbacks/testing.py
```diff
@@ -212,13 +230,42 @@ def on_test_batch_end(
...
     ):
-        return self.on_test_batch_end(
-            trainer, pl_module, outputs, batch, batch_idx, dataloader_idx
+        # test_step returns bilby result object
+        # and optionally a reweighted result
+        result: "AmplfiResult"
+        reweighted: Optional["AmplfiResult"]
+        result, reweighted = outputs
+
+        gpstime = batch[2].cpu().numpy()[0]
```
deepchatterjeeligo, line 251: Is there a nicer way to avoid the index `[2]`? If not, then add a FIXME comment here.

amplfi/train/models/flow.py
```diff
@@ -141,16 +150,40 @@ def validation_step(self, batch, _):
         )
         return loss
 
+    def on_predict_epoch_start(self):
+        self.on_test_epoch_start()
```
deepchatterjeeligo, line 154: what is the difference between predict_epoch and start_epoch in lightning lingo?

amplfi/train/models/flow.py
```diff
@@ -164,55 +197,73 @@ def test_step(self, batch, _) -> AmplfiResult:
...
 
-        # create a dictionary of injection parameters
-        # mapping from parameter string to the true injection value
-        # and add snr if provided
-        injection_parameters = {
-            k: float(v)
-            for k, v in zip(self.inference_params, parameters, strict=False)
-        }
-        injection_parameters["ra"] = injection_parameters["phi"]
+        log_priors = self.training_prior.log_prob(samples_dict)
```
deepchatterjeeligo, line 210: I initially though, this was just log of the prior. Can we use a different variable name, like `log_prob_on_training_prior`? Also add a comment, that this is the posterior samples density estimate on the prior distribution.

amplfi/train/models/flow.py
```diff
@@ -164,55 +197,73 @@ def test_step(self, batch, _) -> AmplfiResult:
...
 
-        return result
+        # optionally reweight to different prior
+        # TODO: include likelihood reweighting
+        reweighted = None
+        if self.target_prior is not None:
+            self._logger.info(
+                f"Reweighting {len(result.posterior)} posterior samples"
+            )
+            reweighted = result.reweight_to_prior(self.target_prior)
```
deepchatterjeeligo, line 246: are these samples, or densities? Please clarify in variable naming.

deepchatterjeeligo: seems like the former, from the logging statement below. In that case, use `reweighted_samples` as the variable name.

amplfi/train/models/flow.py
```diff
@@ -164,55 +197,73 @@ def test_step(self, batch, _) -> AmplfiResult:
...
-            None,
-        )
+    def test_step(self, batch, _) -> AmplfiResult:
+        strain, asds, parameters, snr = batch
+        return self.analyze_event(strain, asds, parameters, snr)
 
-        return result
+    def predict_step(self, batch, _):
+        strain, asds, _ = batch
+        return self.analyze_event(strain, asds, None, None)
```
deepchatterjeeligo, line 261: Very late to ask this basic question: what is the difference between predict and test steps in lightning?

amplfi/train/models/flow.py
```diff
@@ -222,6 +273,8 @@ def cast_as_bilby_result(
             samples:
                 An array of posterior samples of shape
                 (1, num_samples, num_params)
+            log_probs:
+
```
deepchatterjeeligo, line 277: missing docstring for `log_priors` (please rename this to log_probs_prior)

amplfi/utils/result.py
```diff
@@ -112,3 +113,17 @@ def plot_mollview(
...
+    ) -> "AmplfiResult":
+        reweighted = copy(self)
+        keys = list(target_prior.keys())
+        samples_dict = {key: self.posterior[key].values for key in keys}
+        target_probs = target_prior.ln_prob(samples_dict, axis=0)
+        ln_weights = target_probs - self.posterior["log_prior"].values
+        weights = np.exp(ln_weights)
+        reweighted.posterior = bilby.core.result.rejection_sample(
+            reweighted.posterior, weights
+        )
```
deepchatterjeeligo, line 128: Looking at the [implementation in bilby](https://git.ligo.org/lscsoft/bilby/-/blob/master/bilby/core/result.py#L282), a single call _will_ lead to a loss in number of samples, which can be drastic at times. You can call this repeatedly until `max_num_rej_sampling_step` times and collect the samples in the way. Something like
```suggestion
        num_orig_samples = len(self.posterior)
        reweighted_samples = np.empty()
        n_step = 1
        while samples.size <= num_orig_samples and n_step < max_num_rej_sampling_step:
           # call rejection sampling
           # append to reweighted samples
        
```

deepchatterjeeligo (changes requested):
Almost there. Few small improvements on the re-weighting step.

amplfi/train/models/flow.py
EthanMarx: To clarify, this is the prior probability of the sampled posterior samples using the training prior. 

Changed the variable name to `log_prior_of_posterior_samples`

amplfi/train/models/flow.py
EthanMarx: See discussion below

amplfi/train/models/flow.py
EthanMarx: done

PR #248, opened 2025-06-13
add metadata to skymap header

deepchatterjeeligo:


EthanMarx:
@deepchatterjeeligo the latest `ligo.skymap` adds instrument headers to skymaps (see e.g., the skymap on [this](https://ldas-jobs.ligo.caltech.edu/~aframe/o3_mdc_summary_v2/events/event_1435089764/event.html) aframe event page). So would be great if we could also add that to the amplfi skymaps by setting the `instruments` attribute in the table metadata

tests/utils/test_utils.py
```diff
@@ -0,0 +1,54 @@
...
+    assert t.meta["PIXTYPE"] == "HEALPIX"
+    assert t.meta["ORDERING"] == "NUNIQ"
+
+    # check if extra metadata is correctly added
+    skymap_content = skymap.histogram_skymap(
+        ra_samples,
+        dec_samples,
+        dist_samples,
+        nside=n_side,
+        metadata={"INSTRUME": "H1,L1,V1"},
```
deepchatterjeeligo, line 44: The point will be to supply parts of this metadata in the online deployment, like time, and instruments based on the HL vs HLV model.

PR #255, opened 2025-07-17
Add conversion to `ra` from `phi`

EthanMarx:
- When analyzing real events in the predict step, convert to`ra` from `phi`

### DIFF
--- a/amplfi/train/data/datasets/testing.py
+++ b/amplfi/train/data/datasets/testing.py
@@ -449,7 +449,7 @@ class RawStrainTestingDataset(FlowDataset):
 
     def predict_dataloader(self) -> torch.utils.data.DataLoader:
         dataset = torch.utils.data.TensorDataset(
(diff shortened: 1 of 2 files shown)


amplfi/train/data/datasets/testing.py
```diff
@@ -31,6 +31,14 @@ def phi_from_ra(ra: np.ndarray, gpstimes: np.ndarray) -> float:
     return phi
 
 
+def ra_from_phi(phi: np.ndarray, gpstime: float):
+    t = Time(gpstime, format="gps", scale="utc")
+    gmst = t.sidereal_time("mean", "greenwich").to("rad").value
+    ra = phi + gmst
+    ra = ra % (2 * np.pi)
```
deepchatterjeeligo, line 38: I understand it is functionally the same, but can we use np.remainder for consistency with the code right above this. Otherwise looks good.

PR #258, opened 2025-07-31
dynamic versioning using tags and SHAs

deepchatterjeeligo:
This sets up dynamic versioning which is reflected in builds done using say, `uv build`.


deepchatterjeeligo:
@EthanMarx not urgent. I'll do a few tests and then we can review this.

deepchatterjeeligo:
> Can you explain a bit more how this works

Sure. Suppose my latest commit is a tag, then the version is the tag.
```
$ git log -n1 --oneline
96f2ac7 (HEAD, tag: v0.5.3, upstream/main, main) Bump version to 0.5.3 (#257)
$ uv build
Building source distribution...
Building wheel from source distribution...
Successfully built dist/amplfi-0.5.3.tar.gz
Successfully built dist/amplfi-0.5.3-py3-none-any.whl
```
Suppose the latest commit is not a tag but N commits after a tag, then `uv build` will attach a postN to the version. Example, this branch is one commit after 0.5.3, hence `post1`:
```
$ git log -n2 --oneline
8d583e6 (HEAD -> dynamic-versioning, origin/dynamic-versioning) dynamic versioning using tags and SHAs
96f2ac7 (tag: v0.5.3, upstream/main, main) Bump version to 0.5.3 (#257)
$ uv build
Successfully built dist/amplfi-0.5.3.post1.tar.gz
Successfully built dist/amplfi-0.5.3.post1-py3-none-any.whl
```
We can go even one level further and attach the commit SHA if we want.

There are two advantages:
1) We do not need to make any more pull requests to bump the version. In github releases, we just create the release by tagging the relevant commit. E.g. for a hypothetical `0.5.4.`, we will select the commit from `main` like so:
image

2) Our container builds currently carry a stale version number for all commits that are done between two releases. For example, every build between tags 0.5.0 and 0.5.1 current carry the amplfi version 0.5.0. This change will ensure that we clearly demarcate non-released or non-tagged builds. This is good practice now that we are coming close to production.

FWIW I think we should also do the same for aframe.

EthanMarx:
This is great - so we also won't have to constantly bump versions for minor changes as well?

PR #261, opened 2025-08-08
Replace our histogram implementation with `ligo.skymap.healpix_tree.adaptive_healpix_histogram`

EthanMarx:
- Replaces naive histogram skymap with `ligo.skymap.healpix_tree.adaptive_healpix_histogram` followed by `ligo.skymap.bayestar.derasterize`
- Improve plotting (using `ligo.skymap` like code instead of healpy) when plotting skymaps for inspection
- Removes unused `AmplfiResult.calculate_searched_area` (this is done automatically when calling `to_crossmatch_result`)
- Removes some metadata handling, which is automatically done if using `ligo.skymap.io.fits.write_sky_map`

On the Aframe side, we need to migrate to using `ligo.skymap.io.fits.write_sky_map` which will handle the fits metadata for us.


EthanMarx:
Snuck in some other changes here:
- Callbacks now specified in separate config (like the prior) to separate them out
- Added simple integration test of the `cli`

I also kept the ability to use the old histogram method.

PR #263, opened 2025-08-16
Clean up testing datasets

EthanMarx:
- Uses pandas for loading and saving parameters for testing datasets


pyproject.toml
```diff
@@ -69,6 +71,9 @@ dev_template = "{tag}.post{ccount}+git.{sha}"
 requires = [ "setuptools>=77.0.3", "wheel", "setuptools-git-versioning>=2.0,<3",]
 build-backend = "setuptools.build_meta"
 
+[tool.uv.sources]
+ml4gw = { git = "https://github.com/ethanmarx/ml4gw", branch="attention" }
```
deepchatterjeeligo, line 75: this is unrelated change. Otherwise fine.

pyproject.toml
```diff
@@ -28,6 +28,8 @@ dependencies = [
     "s3fs>=2024.12.0,<2025",
     "ligo-skymap>=2.2.2",
     "zuko>=1.4.0",
+    "pandas>=2.3.1",
+    "tables>=3.10.1",
```
deepchatterjeeligo, line 32: I don't see you import tables, at least in the diff. Where is this needed?

deepchatterjeeligo (changes requested):
A couple of questions. Otherwise fine.

pyproject.toml
EthanMarx: Needed for `pandas` hdf support

deepchatterjeeligo: Doesn't pandas provide `pip install pandas[tables]` optional?

deepchatterjeeligo: https://pandas.pydata.org/docs/getting_started/install.html#other-data-sources

It's called hdf5. Let's use that extra.

deepchatterjeeligo (approved):

PR #285, opened 2026-04-11
Heterodyne embedding with decimator

deepchatterjeeligo:
This is just to keep the functionality of decimating the timeseries after heterodyning


bhgupta20 (approved):
Looks good!

PR #286, opened 2026-04-11
Improve model saving

wbenoit26:
Embeds the config and the version in each checkpoint and adds a utility function for more easily loading checkpoints.

### DIFF
--- a/amplfi/utils/checkpoint.py
+++ b/amplfi/utils/checkpoint.py
@@ -1,3 +1,4 @@
+import warnings
 from pathlib import Path
 
 import torch
@@ -5,6 +6,7 @@ import yaml
 from jsonargparse import ArgumentParser
 from ml4gw.transforms import ChannelWiseScaler
 
+import amplfi
 from amplfi.train.architectures.flows import FlowArchitecture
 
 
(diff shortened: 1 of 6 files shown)


amplfi/utils/checkpoint.py
```diff
@@ -0,0 +1,89 @@
...
+    checkpoint = torch.load(
+        checkpoint_path, map_location="cpu", weights_only=False
+    )
+
+    if "amplfi_config" not in checkpoint:
+        raise KeyError(
+            f"Checkpoint '{checkpoint_path}' has no embedded 'amplfi_config'. "
+            "This checkpoint was created before self-contained checkpoints "
+            "were supported."
+        )
```
deepchatterjeeligo, line 42: won't this make the previous checkpoints unusable? How about adding a `error_on_ckpt_mismatch=False` argument and raising this error.

deepchatterjeeligo (changes requested):
I like this @wbenoit26 thanks. Just one question about the KeyError, and the rest looks great.

amplfi/utils/checkpoint.py
wbenoit26: I think any previous checkpoints should be loaded however they were done before. This won't work to load checkpoints without a config.

deepchatterjeeligo: I see. So you are saying that `--ckpt_path` of lightning CLI is not impacted. Where do you intend to use this `from_checkpoint`?

EthanMarx: I think the idea is to make it easier to load in amplfi objects in pure python contexts, as opposed to at the command line

wbenoit26: Exactly, yeah, I should've made that clearer. My own use case is in`buoy`, but it seemed like it might be generally useful to have this utility.

deepchatterjeeligo (approved):

PR #288, opened 2026-05-04
Use numpy >= 2.0

wbenoit26:
I believe the only reason for restricting numpy was compatibility with ml4gw, which now uses numpy >= 2. I've removed this restriction, as well as the explicit dependence on urllib3 as it doesn't seem to be directly used anywhere. I've also updated any deprecated code identified by `ruff check --select NPY`.


deepchatterjeeligo (approved):
Looks good

